PR #1, opened 2021-11-24
Intialize with minimally working code

rahulanand16nov:
This PR adds a minimally working Proxy-Wasm module code to enable testing with Kuadrant controller.
It includes:
- Third-Party protobufs used for communication between the filter and Authorino.
- Example deployment setup that can do end-to-end communication.

~~Right now, the filter just makes a call to the authorino and doesn't act on the response but logs it.~~ There are certain assumptions made that are mentioned in the code (src/ content).


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "authorino"]
+	path = authorino
+	url = https://github.com/Kuadrant/authorino/
```
guicassolato, line 3: Seem a bit overkill that we need to include the entire Authorino repo here. Is this because of the types? I thought we could rely on Envoy protobufs only to integrate the external authorization service.

guicassolato: To deploy Authorino, we should rely on https://github.com/kuadrant/authorino-operator now, BTW.

rahulanand16nov: It's just there to deploy the e2e example like installing CRDs and deploying authorino itself. Yes, we are using Envoy protobufs present in the `vendor-protobufs`.

rahulanand16nov:
Indeed operator is much easier to use! @guicassolato, can you please have a look again?

deploy/wasmplugin.yaml
```diff
@@ -0,0 +1,12 @@
...
+ metadata:
+   name: demo
+   namespace: istio-system
+ spec:
+    selector:
+      matchLabels:
+        app: toystore
+    url: <replace-with-wasm-oci-or-remote-url>
+    pluginConfig:
+      auth_cluster: "outbound|50051||authorino-authorization.default.svc.cluster.local"
```
guicassolato, line 12: When Authorino is deployed by the operator it usually [names the service](https://github.com/Kuadrant/authorino-operator/blob/05cf2ad0c76b6138298f30fbca19da4a2c9b2cf1/controllers/authorino_controller.go#L681) `<name-of-authorino-cr>-authorino-authorization`. This means that the ext authz service would be reachable at `authorino-authorino-authorization.default.svc.cluster.local` for the `Authorino` CR included here.

rahulanand16nov: Thanks for the find! Fixed.

rahulanand16nov:
I have added a script to automatically update protobufs and now filter can handle response as well.

PR #2, opened 2021-12-03
Integrating with limitador

rahulanand16nov:
This PR includes initial implementation to integrate limitador communication into the shim and a new config structure allowing that.

Few points to note in this PR:
- Default failure response has been changed to `500` (Internal Server Error).
- Few action specifier lack implementation like ~header value match~, dynamic metadata, metadata, and extension.
- ~Exclude pattern is not used right now and is optional in the config.~

These points will change as I update the code in the PR but it's best to start reviewing and I'll create new commits for new changes to allow easy reviewing. In the end, those new commits will be squashed into relevant ones.


rahulanand16nov:
Update:
- Exclude pattern is usable now.
- Header Value Match is working except to match against the regex.

rahulanand16nov:
@eguzki This PR is beyond reviewing so can you please approve it?

eguzki (approved):

PR #3, opened 2022-03-03
initiate new API

rahulanand16nov:
This PR implements the new API that is analogous to RateLimitPolicy in the Kuadrant controller but provides a flexible integration point that Envoy's RateLimit filter doesn't for our use cases.

## Verification steps
`cargo test`


src/filter/http_context.rs
```diff
@@ -48,107 +50,36 @@ impl Filter {
...
 
+    fn fetch_request_info(&self) -> RequestInfo {
+        // TODO(rahulanand16nov): Handle error
+        let host_bytes = self.get_property(vec!["request", "host"]).unwrap();
+        let host = String::from_utf8(host_bytes).unwrap();
+
+        let path_bytes = self.get_property(vec!["request", "path"]).unwrap();
+        let path = String::from_utf8(path_bytes).unwrap();
+
+        let method_bytes = self.get_property(vec!["request", "host"]).unwrap();
```
rahulanand16nov, line 61: Aware it should be "method"

rahulanand16nov:
Added test to check if deserializing of config. It also gives us a sample config to showcase what is required.

Tests are executed using `cargo test`

rahulanand16nov:
cc @maleck13

maleck13 (commented):
Not sure I can really review this as have no experience with Rust. One thing I think that would be a good addition would be to update the README with details of what it does and the config it now takes

eguzki (approved):
I cannot review this as it should be done. 

One day, we need to review all this code in pair programming session

PR #14, opened 2022-06-20
policy index

eguzki:
The WASM-SHIM builds a tree based data structure holding the rate limit policies. The longest hostname match is used to select the policy to be applied. Only one policy is being applied per invocation.

It is based on using https://docs.rs/radix_trie/latest/radix_trie/struct.Trie.html, which implements longest prefix match. So, for longest *suffix* match (for subdomain match), reverse operation is needed.

Note: this PR target is `new-design` branch and more PRs are coming

### DIFF
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -38,6 +38,22 @@ pub struct RateLimitPolicy {
 }
 
 impl RateLimitPolicy {
+    pub fn new(
+        hosts: GlobPatternSet,
+        rules: Option<Vec<Rule>>,
+        global_actions: Option<Vec<RLA_action_specifier>>,
+        upstream_cluster: Option<String>,
+        domain: Option<String>,
+    ) -> Self {
+        RateLimitPolicy {
+            hosts,
+            rules,
+            global_actions,
+            upstream_cluster,
+            domain,
+        }
+    }
+
     pub fn rules(&self) -> Option<&[Rule]> {
         self.rules.as_deref()
     }
--- a/src/policy_index.rs
+++ b/src/policy_index.rs
@@ -1,0 +1,122 @@
+use radix_trie::Trie;
+
+use crate::configuration::RateLimitPolicy;
+
+pub struct PolicyIndex {
+    raw_tree: Trie<String, RateLimitPolicy>,
+}
+
+impl PolicyIndex {
+    pub fn new() -> Self {
+        Self {
+            raw_tree: Trie::new(),
+        }
+    }
+
+    pub fn insert(&mut self, subdomain: &str, policy: RateLimitPolicy) {
+        let rev = Self::reverse_subdomain(subdomain);
+        self.raw_tree.insert(rev, policy);
+    }
+
+    pub fn get_longest_match_policy(&self, subdomain: &str) -> Option<&RateLimitPolicy> {
+        let rev = Self::reverse_subdomain(subdomain);
+        self.raw_tree.get_ancestor_value(&rev)
+    }
+
+    fn reverse_subdomain(subdomain: &str) -> String {
+        let mut s = subdomain.to_string();
+        s.push('.');
+        if s.chars().nth(0).unwrap() == '*' {
+            s.remove(0);
+        } else {
+            s.insert(0, '$'); // $ is not a valid domain char
+        }
+        s.chars().rev().collect()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::configuration::RateLimitPolicy;
+    use crate::glob::GlobPatternSet;
+    use crate::policy_index::PolicyIndex;
+
+    fn build_ratelimit_policy(domain: String) -> RateLimitPolicy {
+        RateLimitPolicy::new(GlobPatternSet::default(), None, None, None, Some(domain))
+    }
+
+    #[test]
+    fn not_wildcard_subdomain() {
+        let mut index = PolicyIndex::new();
+        let rlp1 = build_ratelimit_policy(String::from("example.com"));
+        index.insert("example.com", rlp1);
+
+        let val = index.get_longest_match_policy("test.example.com");
+        assert_eq!(val.is_none(), true);
+
+        let val = index.get_longest_match_policy("other.com");
+        assert_eq!(val.is_none(), true);
+
+        let val = index.get_longest_match_policy("net");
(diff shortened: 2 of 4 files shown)


src/policy_index.rs
```diff
@@ -0,0 +1,122 @@
+use radix_trie::Trie;
+
+use crate::configuration::RateLimitPolicy;
+
+pub struct PolicyIndex {
+    raw_tree: Trie<String, RateLimitPolicy>,
```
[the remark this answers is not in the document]
eguzki: In such case, it is the responsability of the kuadrant control plane to consolidate into one RLP merging those. 

That happens when Gateway RLP targets `*.com` and some HTTPRoute targets `*.example.com`.

Makes sense that the wasm-shim expects one RLP per (exact) subdomain?

src/policy_index.rs
```diff
@@ -0,0 +1,122 @@
...
+            raw_tree: Trie::new(),
+        }
+    }
+
+    pub fn insert(&mut self, subdomain: &str, policy: RateLimitPolicy) {
+        let rev = Self::reverse_subdomain(subdomain);
+        self.raw_tree.insert(rev, policy);
+    }
+
+    pub fn get_longest_match_policy(&self, subdomain: &str) -> Option<&RateLimitPolicy> {
```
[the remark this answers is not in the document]
eguzki: I think I answered in https://github.com/Kuadrant/wasm-shim/pull/14#discussion_r902534863

Let me know if not

src/policy_index.rs
```diff
@@ -0,0 +1,122 @@
...
+    use crate::policy_index::PolicyIndex;
+
+    fn build_ratelimit_policy(domain: String) -> RateLimitPolicy {
+        RateLimitPolicy::new(GlobPatternSet::default(), None, None, None, Some(domain))
+    }
+
+    #[test]
+    fn not_wildcard_subdomain() {
+        let mut index = PolicyIndex::new();
+        let rlp1 = build_ratelimit_policy(String::from("example.com"));
```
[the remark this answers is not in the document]
eguzki: in this test, this is like an ID of the RLP. Thus, when 
```
index.get_longest_match_policy(domain)
```
returns a RLP, the test makes sure it is the expected one.  

The real "indexing" value is given in the `insert` method

```
index.insert("example.com", rlp1);
```

src/configuration.rs
```diff
@@ -38,6 +38,22 @@ pub struct RateLimitPolicy {
 }
 
 impl RateLimitPolicy {
+    pub fn new(
+        hosts: GlobPatternSet,
+        rules: Option<Vec<Rule>>,
+        global_actions: Option<Vec<RLA_action_specifier>>,
+        upstream_cluster: Option<String>,
+        domain: Option<String>,
```
alexsnaps, line 46: Unsure about these `Option`s here, but I guess we'll see how this pans out.

src/policy_index.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::configuration::RateLimitPolicy;
+    use crate::glob::GlobPatternSet;
+    use crate::policy_index::PolicyIndex;
+
+    fn build_ratelimit_policy(domain: String) -> RateLimitPolicy {
```
alexsnaps, line 44: Just for readability I maybe would change this signature to take `&str`, so that call sites are "nicer" to read in the tests and make it a `String` in this function.

src/policy_index.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    #[test]
+    fn not_wildcard_subdomain() {
+        let mut index = PolicyIndex::new();
+        let rlp1 = build_ratelimit_policy(String::from("example.com"));
+        index.insert("example.com", rlp1);
+
+        let val = index.get_longest_match_policy("test.example.com");
+        assert_eq!(val.is_none(), true);
```
alexsnaps, line 55: `assert!`? but I either is fine with me…

src/policy_index.rs
```diff
@@ -0,0 +1,122 @@
...
+        let val = index.get_longest_match_policy("other.com");
+        assert_eq!(val.is_none(), true);
+
+        let val = index.get_longest_match_policy("net");
+        assert_eq!(val.is_none(), true);
+
+        let val = index.get_longest_match_policy("example.com");
+        assert_eq!(val.is_some(), true);
+        assert_eq!(val.unwrap().domain().is_some(), true);
+        assert_eq!(val.unwrap().domain().unwrap(), "example.com");
```
alexsnaps, line 66: I like `assert_eq!(val.unwrap().domain(), Some("example.com"))`, but again… Gives these nice messages:
```
'assertion failed: `(left == right)`
  left: `None`,
 right: `Some("example.com")`
```

alexsnaps (approved):

src/policy_index.rs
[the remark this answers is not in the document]
eguzki: changing it for readability

src/policy_index.rs
eguzki: :facepalm:

src/policy_index.rs
eguzki: done

src/policy_index.rs
eguzki: done

src/policy_index.rs
eguzki: done

src/configuration.rs
rahulanand16nov: Without `Option`, you make a strict requirement for that field to be present else, it will fail the shim during loading, and the control plane has no option to get that info. Thus, always avoid irrecoverable failure in wasm :)

rahulanand16nov: yeah, a default is a much better choice like `NO_DOMAIN_FOUND`.

src/policy_index.rs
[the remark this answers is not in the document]
rahulanand16nov: If it's something that can be done in the control-plane then fine by me.

src/configuration.rs
eguzki: a default domain can lead to unexpected rate limit behavior. That's why I tend to think that it should be required and fail if not present.

We can leave as it is for now.

PR #17, opened 2022-06-21
Gh action rust

eguzki:
GH Action to run

* cargo check
* cargo test
* cargo fmt
* cargo clippy

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -1,0 +1,91 @@
+---
+name: Rust
+
+on:
+  push:
+    branches:
+      - '*'
+  pull_request:
+    branches:
+      - '*'
+  schedule:
+    - cron: '5 4 * * *'
+
+jobs:
+  check:
+    name: Check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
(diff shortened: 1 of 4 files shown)


.github/workflows/rust.yaml
```diff
@@ -0,0 +1,91 @@
...
+          profile: minimal
+          toolchain: stable
+          override: true
+      - uses: arduino/setup-protoc@v1
+        with:
+          version: '3.x'
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --all-features
```
eguzki, line 49: When `wasm32-unknown-unknown` target is added:

```
cargo test --lib --target wasm32-unknown-unknown --release
--- bunch of warning ---
    Finished release [optimized] target(s) in 0.04s
     Running unittests (target/wasm32-unknown-unknown/release/deps/wasm_shim-b1ac92665724d8ed.wasm)
/home/eguzki/git/wasm-shim2/target/wasm32-unknown-unknown/release/deps/wasm_shim-b1ac92665724d8ed.wasm: 1: Syntax error: end of file unexpected
error: test failed, to rerun pass '--lib'
```

rahulanand16nov: make sense since `.wasm` can only run in a sandbox but `cargo test` should also be fine for unit tests.

rahulanand16nov (approved):
LGTM

PR #19, opened 2022-06-27
New plugin config struct

eguzki:
### What

New PluginConfiguration structure. 

Each rate limit policy has a logical name and a set of hostnames to activate it based on the incoming request’s host header.

The policy contains a list of configurations to build a list of Envoy's RLS descriptors. 

Each configuration has associated, optionally, a set of rules to match. Rules allow matching `hosts` and/or `methods` and/or `paths`. 

Each configuration object defines a list of actions to produce descriptor entries.

Example of plugin configuration object

```yaml
failure_mode_deny: true
rate_limit_policies:
  - name: toystore
    hostnames: ["*.toystore.com"]
    gateway_actions:
      - rules:
          - paths: ["/admin/toy"]
            methods: ["GET"]
            hosts: ["pets.toystore.com"]
        configurations:
          - actions:
            - generic_key:
                descriptor_key: admin
                descriptor_value: "1"
```

For the context, the replaced plugin configuration object:

```yaml
#  The filter’s behaviour in case the rate limiting service does not respond back. When it is set to true, Envoy will not allow traffic in case of communication failure between rate limiting service and the proxy.
failure_mode_deny: true 
ratelimitpolicies:
  default/toystore: # rate limit policy {NAMESPACE/NAME}
    hosts: # HTTPRoute hostnames
      - '*.toystore.com'
    rules: # route level actions
      - operations:
          - paths:
              - /admin/toy
            methods:
              - POST
              - DELETE
        actions:
          - generic_key:
              descriptor_value: yes
              descriptor_key: admin
    global_actions: # virtualHost level actions
      - generic_key:
          descriptor_value: yes
          descriptor_key: vhaction
    upstream_cluster: rate-limit-cluster # Limitador address reference
    domain: toystore-app # RLS protocol domain value
```

Update highlights:
* [*minor*] rate_limit_policies is a list instead of a map indexed by the name/namespace.
* [*major*] no distinction between "rules" and global actions
* [*major*] more aligned with RLS: multiple descriptors structured by "rate limit configurations" with matching rules https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-virtualhost-rate-limits

Note: It is expected that the code does not compile and tests do not pass. This PR tries to include only the minimum changes for the struct update. New PR's coming soon to add completeness.

### DIFF
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -1,77 +1,80 @@
 use crate::envoy::RLA_action_specifier;
-use crate::glob::GlobPatternSet;
+use crate::policy_index::PolicyIndex;
 use serde::Deserialize;
-use std::collections::HashMap;
-
-#[derive(Deserialize, Debug, Clone)]
-pub struct Operation {
-    #[serde(default)]
-    pub paths: GlobPatternSet,
-    #[serde(default)]
-    pub methods: GlobPatternSet,
-}
 
 #[derive(Deserialize, Debug, Clone)]
 pub struct Rule {
-    pub operations: Option<Vec<Operation>>,
+    pub paths: Option<Vec<String>>,
+    pub hosts: Option<Vec<String>>,
+    pub methods: Option<Vec<String>>,
+}
+
+#[derive(Deserialize, Debug, Clone)]
+pub struct Configuration {
     pub actions: Option<Vec<RLA_action_specifier>>,
 }
 
-impl Rule {
-    pub fn operations(&self) -> Option<&[Operation]> {
-        self.operations.as_deref()
-    }
-
-    pub fn actions(&self) -> Option<&[RLA_action_specifier]> {
-        self.actions.as_deref()
-    }
+#[derive(Deserialize, Debug, Clone)]
+pub struct GatewayAction {
+    pub rules: Option<Vec<Rule>>,
+    pub configurations: Option<Vec<Configuration>>,
 }
 
 #[derive(Deserialize, Debug, Clone)]
 pub struct RateLimitPolicy {
-    #[serde(default)]
-    hosts: GlobPatternSet,
-    rules: Option<Vec<Rule>>,
-    global_actions: Option<Vec<RLA_action_specifier>>,
-    upstream_cluster: Option<String>,
-    domain: Option<String>,
+    pub name: String,
+    pub rate_limit_domain: String,
+    pub upstream_cluster: String,
+    pub hostnames: Vec<String>,
+    pub gateway_actions: Vec<GatewayAction>,
 }
 
 impl RateLimitPolicy {
+    #[cfg(test)]
     pub fn new(
-        hosts: GlobPatternSet,
-        rules: Option<Vec<Rule>>,
-        global_actions: Option<Vec<RLA_action_specifier>>,
-        upstream_cluster: Option<String>,
-        domain: Option<String>,
+        name: String,
+        rate_limit_domain: String,
+        upstream_cluster: String,
+        hostnames: Vec<String>,
+        gateway_actions: Vec<GatewayAction>,
     ) -> Self {
         RateLimitPolicy {
-            hosts,
-            rules,
-            global_actions,
+            name,
+            rate_limit_domain,
             upstream_cluster,
-            domain,
+            hostnames,
+            gateway_actions,
+        }
+    }
+}
+
+pub struct FilterConfig {
+    pub index: PolicyIndex,
+    // Deny request when faced with an irrecoverable failure.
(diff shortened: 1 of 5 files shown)


rahulanand16nov:
@eguzki What's the reasoning behind this change?

eguzki:
> @eguzki What's the reasoning behind this change?

It's in the description. Let me know if it is not clear enough.

rahulanand16nov:
> It's in the description. Let me know if it is not clear enough.

Sorry Eguzki, but it's just describing what config is supposed to contain but not why the new one is better than the last one?

eguzki:
* [*minor*] `rate_limit_policies` is a list instead of a map indexed by the name/namespace. 
* [*major*] no distinction between "rules" and global actions
* [*major*] more aligned with RLS: multiple descriptors structured by "rate limit configurations" with matching rules https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-virtualhost-rate-limits

alexsnaps:
Also is something missing here? Build as broken for me locally… Some mix up in the config types expected?

eguzki:
> Also is something missing here? Build as broken for me locally… Some mix up in the config types expected?

As description says, `It is expected that the code does not compile and tests do not pass. This PR tries to include only the minimum changes for the struct update. New PR's coming soon to add completeness.`

This PR is just to validate the new struct schema. Also note that the PR does not target `main` branch. It is part of a bigger work piece

src/configuration.rs
```diff
@@ -1,77 +1,80 @@
...
-    hosts: GlobPatternSet,
-    rules: Option<Vec<Rule>>,
-    global_actions: Option<Vec<RLA_action_specifier>>,
-    upstream_cluster: Option<String>,
-    domain: Option<String>,
+    pub name: String,
+    pub rate_limit_domain: String,
+    pub upstream_cluster: String,
+    pub hostnames: Vec<String>,
+    pub gateway_actions: Vec<GatewayAction>,
```
eguzki, line 29: not really happy with the name `gateway_actions`.

Maybe `rate_limit_configurations`? Sometimes bad naming indicates bad structure design. If you cannot name samething, maybe it should not be there.

Dropping here the entire struct

```yaml
failure_mode_deny: true
rate_limit_policies:
  - name: toystore
    hostnames: ["*.toystore.com"]
    gateway_actions:
      - rules:
          - paths: ["/admin/toy"]
            methods: ["GET"]
            hosts: ["pets.toystore.com"]
        configurations:
          - actions:
            - generic_key:
                descriptor_key: admin
                descriptor_value: "1"
```

Currently `gateway_actions` is a list of "configuration lists" and each configuration list is conditionally applied by the rule matching. 

I was tempted to add the `rules` inside the `configuration` object, sibling of `actions`. Influenced by Envoy's rate limiting design, I decided to go as it is now. Envoy's rate limiting design adds the configuration list (**not only one configuration**) at the route level. Check out envoy's doc at  [RouteAction](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-routeaction). Which is equivalent to say 
```yaml
rules:
  - paths: ["/some/path"]
    methods: ["GET"]
configurations:
  - actions:
    -  generic_key:
          descriptor_key: admin
          descriptor_value: "1"
```

The important bit I want to highlight is that, in envoy, for a given route, you can define a list of configurations, each configuration potentially (depending on the request context) yielding one rate limit descriptor.  So each configuration yields 0 or 1 rate limit descriptor. Thus, each route can, potentially, generate multiple descriptors. 

Equivalence table:

Kuadrant object | Envoys object
--- | --- 
GatewayAction struct | [config.route.v3.RouteAction](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-routeaction)
Configuration struct| [config.route.v3.RateLimit](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-ratelimit)

So, each gateway action object (envoys RouteAction object) holds a list of configuration objects (a list of envoy's RateLimit objects). 

While envoy's design has rules implicit because RouteAction is defined at specific http route (method and path) in the virtualhost object, kuadrant needs to explicitly define this rules.  

Any input would be appreciated

rahulanand16nov: `rate_limit_configs` sounds better :)

rahulanand16nov (approved):
LGTM.

PR #20, opened 2022-06-27
http context implementation with the new filter config struct

eguzki:
This PR depends on the new struct defined in #19

### What

New behavior defined.

The WASM-SHIM builds a tree based data structure holding the rate limit policies. The longest (sub)domain match is used to select the policy to be applied. Only one policy is being applied per invocation.

Policy configurations are defined at
```
rate_limit_policies[*].gatewat_actions[*].configurations
```
For example:
```yaml
configurations:
- actions:
   - generic_key:
        descriptor_key: admin
        descriptor_value: "1"
```

Each configuration produces, at most, one descriptor. Depending on the incoming request, one configuration may or may not produce a rate limit descriptor.

Each policy configuration has associated, optionally, a set of rules to match. Rules allow matching hosts and/or methods and/or paths. Matching occurs when at least one rule applies against the incoming request. If rules are not set, it is equivalent to matching all the requests.

For example:

```yaml
gateway_actions:
      - rules:
          - paths: ["/admin/toy"]
            methods: ["GET"]
            hosts: ["pets.toystore.com"]
        configurations:
          - actions:
            - generic_key:
                descriptor_key: admin
                descriptor_value: "1"
```

Each configuration object defines a list of actions. Each action may (or may not) produce a descriptor entry (descriptor list item).
If an action cannot append a descriptor entry, no descriptor is generated for the configuration.

The external rate limit service will be called when there is at least one not empty descriptor generated by the list of configurations.


src/filter/http_context.rs
```diff
@@ -1,159 +1,259 @@
...
+        }
     }
-}
 
-impl HttpContext for Filter {
-    fn on_http_request_headers(&mut self, _: usize, _: bool) -> Action {
-        let context_id = self.context_id();
-        info!("context #{}: on_http_request_headers called", context_id);
+    fn apply_rule(&self, rule: &Rule) -> bool {
+        if let Some(paths) = &rule.paths {
```
alexsnaps, line 106: So that was some of my reservation around the `Option` here… why not have an empty vec when nothing is there and have the field be `paths: Vec<String>`? Tho maybe it make the whole check slightly more awkward… arguably, now `Some(empty vec) != None`…

src/filter/http_context.rs
```diff
@@ -1,159 +1,259 @@
...
 pub struct Filter {
     pub context_id: u32,
     pub config: Rc<FilterConfig>,
 }
 
 impl Filter {
-    fn config(&self) -> &FilterConfig {
-        &self.config
+    fn request_path(&self) -> String {
+        // Note(eastizle): self.get_http_request_header(":path") is equivalent?
```
alexsnaps, line 25: Worthwhile having a look at, the former would be nice… avoids some work from the `vec!`

src/filter/http_context.rs
```diff
@@ -1,159 +1,259 @@
...
+            Some(rules) => rules.iter().any(|rule| self.apply_rule(rule)),
+        }
     }
-}
 
-impl HttpContext for Filter {
-    fn on_http_request_headers(&mut self, _: usize, _: bool) -> Action {
-        let context_id = self.context_id();
-        info!("context #{}: on_http_request_headers called", context_id);
+    fn apply_rule(&self, rule: &Rule) -> bool {
```
alexsnaps, line 105: nit: `rule_applies` ?

src/filter/http_context.rs
```diff
@@ -1,159 +1,259 @@
...
+            .flat_map(|configuration| &configuration.actions)
+            // filter None and return descriptors
+            .flat_map(|actions| self.build_single_descriptor(actions))
+            .collect()
+    }
 
-        Ok(rl_req)
+    fn filter_configurations_by_rules(&self, rules_opt: &Option<Vec<Rule>>) -> bool {
+        match rules_opt {
+            // no rules is equivalent to matching all the requests.
```
alexsnaps, line 99: This bits confuses me… (see below)
But relatively the same point of why we warp `Vec`s in `Option`? I feel I'm missing something. Especially that, again if I read this right, `Vec.is_empty() != None`… Or I'm just too tired and reading this all wrong ¯\_(°_o)_/¯

alexsnaps (commented):
I might need another pass at this, I was reviewing the entire changesets locally… and I might just be wrong about things here… but I definitively miss something

src/filter/http_context.rs
eguzki: I need to play a little bit more with serde, but I guess `Option` is needed to differentiate between:
* `rules` not defined
```yaml
gateway_actions:
- configurations: {}
```
* rules defined but empty
 ```yaml
gateway_actions:
- rules: []
  configurations: {}
```

File: src/filter/http_context.rs [hunk 5]
```diff
@@ -1,159 +1,259 @@
...
+            None => true,
+            Some(rules) => rules.iter().any(|rule| self.apply_rule(rule)),
```
eguzki, line 101: TIL: when applying `any` to an empty iterator, it returns false. :facepalm: 

So empty vector would return `false` and it should return `true`

src/filter/http_context.rs
alexsnaps: Are those two equivalent tho? They are right?

eguzki: For the kuadrant's purposes, they are equivalent. I have not checked (yet) if the de-serialization works the same if Option is removed. My guess is that it does not work the same.

eguzki: Well. that's up to us. 

if I define
```yaml
- rules: []
```
What is expected? the spec says at least one set of rules must match. 

```
Matching occurs when at least one rule applies against the incoming request. If rules are not set, it is equivalent to matching all the requests.
```
Maybe the current implementation is right. If no rules are set, then the rate limit config is applied. If empty rules is set, then none applies, therefore the rate limit configuration should not apply.

wdyt?

maleck13: Hmm my initial instinct when reading this is that ``` - rules: [] ``` is equiv to not specifying rules. At least that is the way I read those two examples.

src/filter/http_context.rs
```diff
@@ -1,159 +1,259 @@
...
+        // Note(eastizle): self.get_http_request_header(":path") is equivalent?
+        let path_bytes = self.get_property(vec!["request", "path"]).unwrap();
+        String::from_utf8(path_bytes).unwrap()
     }
 
-    fn context_id(&self) -> u32 {
-        self.context_id
+    fn request_method(&self) -> String {
+        // Note(eastizle): self.get_http_request_header(":method") is equivalent?
+        let method_bytes = self.get_property(vec!["request", "method"]).unwrap();
```
rahulanand16nov, line 32: can use `get_http_request_header` here as well

src/filter/http_context.rs
eguzki: for `rules`, `[]` == `nil` meaning that rate limit configurations will apply unconditionally. I think that `Option` will not be needed. 

Only when there is at least one rule, matching occurs when at least one rule applies

eguzki: implemented

eguzki:
ready for a new review @alexsnaps @rahulanand16nov

rahulanand16nov (approved):
LGTM

Issue #10, opened 2022-05-30
RLP Support Meta Data actions


maleck13:
- [x] Missing envoy rate limiting actions required to implement authenticated rate limit 
* https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-ratelimit-action
* Attributes available at the wasm shim: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes

- [ ] Update user guide to include authenticated rate limit

#### dev notes
* `descriptor_from_actions` in [utils.rs](https://github.com/Kuadrant/wasm-shim/blob/main/src/utils.rs)

eguzki:
Related issue: https://github.com/Kuadrant/kuadrant-controller/issues/144

eguzki:
@rahulanand16nov  I wonder if we can close this as it was implemented in #22

rahulanand16nov:
I was actually waiting for https://github.com/Kuadrant/kuadrant-controller/pull/184 but it's merged now. Closed :)

PR #22, opened 2022-07-14
Metadata action type

rahulanand16nov:
#10 

Implements metadata acton type along with some unit tests.

I have not done the end-to-end test yet because changes are required in the Kuadrant controller before we test the user-id-based rate limit.

### DIFF
--- a/deploy/Makefile
+++ b/deploy/Makefile
@@ -2,7 +2,7 @@ MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 PROJECT_PATH := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
 IMAGE_VERSION ?= latest
 DOCKER ?= podman
-IMAGE_REPOSITORY ?= rahanand
+IMAGE_REPOSITORY ?= kuadrant
 
 .PHONY: build
 build-image:
--- a/deploy/Makefile
+++ b/deploy/Makefile
@@ -2,15 +2,16 @@ MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 PROJECT_PATH := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
 IMAGE_VERSION ?= latest
 DOCKER ?= podman
+IMAGE_REPOSITORY ?= rahanand
 
 .PHONY: build
-build:
-	$(DOCKER) build -t quay.io/rahanand/wasm-shim:$(IMAGE_VERSION) -f $(PROJECT_PATH)/Dockerfile
+build-image:
+	$(DOCKER) build -t quay.io/$(IMAGE_REPOSITORY)/wasm-shim:$(IMAGE_VERSION) -f $(PROJECT_PATH)/Dockerfile
 
 .PHONY: clean
-clean:
-	$(DOCKER) rmi quay.io/rahanand/wasm-shim:$(IMAGE_VERSION)
+clean-image:
+	$(DOCKER) rmi quay.io/$(IMAGE_REPOSITORY)/wasm-shim:$(IMAGE_VERSION)
 
 .PHONY: push
-push:
-	$(DOCKER) push quay.io/rahanand/wasm-shim:$(IMAGE_VERSION)
+push-image:
+	$(DOCKER) push quay.io/$(IMAGE_REPOSITORY)/wasm-shim:$(IMAGE_VERSION)
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -124,6 +124,28 @@ mod test {
                             "descriptor_key": "admin",
                             "descriptor_value": "1"
                         }
+                    },
+                    {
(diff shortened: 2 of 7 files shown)


src/filter/http_context.rs
```diff
@@ -228,7 +228,36 @@ impl Filter {
...
+                        None => {
+                            debug!("metadata key not found");
+                            descriptor_entry.set_key(default_value.into());
+                        }
+                        Some(metadata_bytes) => match String::from_utf8(metadata_bytes) {
+                            Err(e) => {
+                                debug!("failed to parse metadata value: {}", e);
+                                descriptor_entry.set_value(default_value.into());
+                            }
+                            Ok(metadata_value) => descriptor_entry.set_value(metadata_value),
```
eguzki, line 257: `descriptor_entry` needs to be added to the `entries`
```rust
entries.push(descriptor_entry);
```

src/filter/http_context.rs
```diff
@@ -228,7 +228,36 @@ impl Filter {
...
+                RLA_action_specifier::metadata(md) => {
+                    // Note(rahul): defaulting to dynamic metadata source right now.
+                    let metadata_key = &md.get_metadata_key().key;
+                    let mut metadata_path: Vec<&str> = md
+                        .get_metadata_key()
+                        .get_path()
+                        .iter()
+                        .map(|path_segment| path_segment.get_key())
+                        .collect();
+                    let default_value = md.get_default_value();
```
eguzki, line 240: default value is a required fields? in the envoy docs, it is not https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-ratelimit-action-metadata

Maybe it is in the protoc objects.

What if the default value is not set?

```
default_value

    ([string](https://developers.google.com/protocol-buffers/docs/proto#scalar)) An optional value to use if metadata_key is empty. If not set and no value is present under the metadata_key then no descriptor is generated.
```

eguzki (changes requested):

eguzki:
> I have not done the end-to-end test yet because changes are required in the Kuadrant controller before we test the user-id-based rate limit.

You can test locally with the docker compose env

src/filter/http_context.rs
rahulanand16nov: It would use the default `String` value i.e. `""`

src/filter/http_context.rs
rahulanand16nov: fixed

src/configuration.rs
```diff
@@ -141,6 +163,35 @@ mod test {
...
+        let configurations = &gateway_actions[0].configurations;
+        assert_eq!(configurations.len(), 1);
+
+        let actions = &configurations[0].actions;
+        assert_eq!(actions.len(), 2);
+        assert!(std::matches!(
+            actions[0],
+            RLA_action_specifier::generic_key(_)
+        ));
+        assert!(std::matches!(actions[1], RLA_action_specifier::metadata(_)));
```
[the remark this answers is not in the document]
rahulanand16nov: fixed

deploy/Makefile
```diff
@@ -2,15 +2,16 @@ MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 PROJECT_PATH := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
 IMAGE_VERSION ?= latest
 DOCKER ?= podman
+IMAGE_REPOSITORY ?= rahanand
```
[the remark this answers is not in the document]
rahulanand16nov: fixed

src/filter/http_context.rs
rahulanand16nov: Looks like it will fail the parsing of configuration and yes, it's due to protoc objects.

rahulanand16nov:
> > I have not done the end-to-end test yet because changes are required in the Kuadrant controller before we test the user-id-based rate limit.
> 
> You can test locally with the docker compose env. No need for the kuadrant's env to test it.

I need some way to set the dynamic metadata which is done by Authorino. :thinking:

src/filter/http_context.rs
rahulanand16nov: I wonder if `default_value` is empty then we should skip this action altogether.

rahulanand16nov:
@eguzki I have used the `header-to-metadata` filter and updated the local development configuration to use metadata action. LGTM.

src/filter/http_context.rs
eguzki: I would say so

rahulanand16nov: I think a better option is to let RLP validation tell the user that the default value is not provided.

rahulanand16nov: Done.

rahulanand16nov:
@eguzki Okay to merge now?

eguzki (approved):

PR #24, opened 2022-09-09
prefix based path matching rule

eguzki:
### what

Support prefix based path matching

The following config will generate descriptors for the request `GET /v1/service/my-service/something`

```yaml
failure_mode_deny: true
rate_limit_policies:
  - name: toystore
    hostnames: ["*.toystore.com"]
    gateway_actions:
      - rules:
          - paths: ["/v1/service*"]   <- prefix based path matching rule
            methods: ["GET"]
        configurations:
          - actions:
            - generic_key:
                descriptor_key: admin
                descriptor_value: "1"
```
Required by https://github.com/Kuadrant/kuadrant-controller/issues/202

Additionally:
* use `latest` limitador version in the dev environment (based on docker compose)
* Adapt dev environment limits file to the latest syntax
* FIX metadata action type when the key is not found

### verification steps

Run dev env

```
make development
```
The wasm config in the envoy for the `*.b.com` workload look like this
```json
{
  "name": "rlp-b",
  "rate_limit_domain": "rls-domain-b",
  "upstream_cluster": "limitador",
  "hostnames": ["*.b.com"],
  "gateway_actions": [
    {
      "rules": [
        {
          "paths": ["/not/exising/path", "/any*"]           <- prefix based path matching
        }
      ],
      "configurations": [
        {
          "actions": [
            {
              "generic_key": {
                "descriptor_key": "some_generic_key",
                "descriptor_value": "some_generic_value"
              }
            }
          ]
        }
      ]
    }
  ]
}
```

Check that the request 
```
curl -H "Host: test.b.com" http://127.0.0.1:18000/anything
```
is generating descriptor entries. It can easily be seen in the lmitador's logs or checking that after 5 request, the requests are rate limited with 429 responses. If it is rate limited, descriptors must have been generated. Currently the limit file looks like this

```yaml
 - namespace: rls-domain-b                         
   max_value: 5                                    
   seconds: 10                                     
   conditions:                                     
   ⦙ - "some_generic_key == 'some_generic_value'"  
```


eguzki:
kind reminder for review @Kuadrant/engineering

src/utils.rs
```diff
@@ -113,6 +113,16 @@ pub fn subdomain_match(subdomain: &str, authority: &str) -> bool {
     authority.ends_with(subdomain.replace('*', "").as_str())
 }
 
+pub fn path_match(path_pattern: &str, request_path: &str) -> bool {
+    if path_pattern.ends_with('*') {
+        let mut cp = path_pattern.to_string();
+        cp.pop();
```
alexsnaps, line 119: The `to_string()` here isn't really necessary, you want to get rid of the last char, you can keep it as a slice and just narrow this slice: `&path_pattern[..path_pattern.len() - 1]` which returns a slice…

alexsnaps (approved):

src/utils.rs
alexsnaps: … which saves one heap allocation, this points to the same `str` wherever it sits…

src/utils.rs
```diff
@@ -124,4 +132,17 @@ mod tests {
...
+    #[test]
+    fn path_match() {
+        assert!(utils::path_match("/cats", "/cats"));
+        assert!(utils::path_match("/", "/"));
+        assert!(utils::path_match("/*", "/"));
+        assert!(utils::path_match("/*", "/cats/something"));
+        assert!(utils::path_match("/cats/*", "/cats/"));
+        assert!(utils::path_match("/cats/*", "/cats/something"));
+        assert!(!utils::path_match("/cats/*", "/cats"));
+        assert!(utils::path_match("/cats*", "/catsanddogs"));
```
alexsnaps, line 145: I was actually surprised by this… because of [this you said](https://kuadrant.slack.com/archives/C02PB55TERE/p1662736074703149):

>TIL that the prefix path match definition of Gateway API is not what I would expect. https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPRouteMatch
>/abc matches /abc , /abc/ , /abc/cde, but does not match /abcd :exploding_head:

eguzki: well, kuadrant does not implement segment path prefix matching, just strict prefix matching. 

Should we implement in the same way?

alexsnaps: That's a good question… that i don't plan on answering on this PR! 
I'd merge just merge that, but I'd like to wrap my head around what it all means moving forward… Tho, we can always do some replacement at "higher" abstractions that'd consider `/abc*` segment path prefix matching (if that's how its called), but as `/abc/*` for that layer of the stack… So I don't feel that the cost of this turning out be being wrong would be huge… So remains a 👍 from me!

PR #27, opened 2022-12-20
multiarch CI build

eguzki:


alexsnaps:
Confused... the resulting image only has wasm in it... why do we need a multi arch?

eguzki:
you are right, no need for this.

PR #36, opened 2023-06-15
e2e tests

alexsnaps:
And uses the work based of [this](https://github.com/proxy-wasm/test-framework/), but with [the changes](https://github.com/proxy-wasm/test-framework/compare/master...alexsnaps:test-framework:kuadrant) required to run ABI v0.2.1 and grpc stuff 

I propose: 
 - [x] Forking proxy-wasm/test-framework into our Kuadrant org
 - [x] Push kuadrant branch from my fork to that and make that the "main" branch
 - [x] (optional) Would we rather want to have cargo declare the dependency on that branch on github, or keep it as submodule? The latter makes it easier to alter as needed, when needed; the former makes the build scripts et al tidier…

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,3 +4,28 @@ mod filter;
 mod glob;
 mod policy_index;
 mod utils;
+
+#[cfg(test)]
+mod tests {
+    use crate::envoy::{HeaderValue, RateLimitResponse, RateLimitResponse_Code};
+    use protobuf::Message;
+
+    #[test]
+    fn grpc() {
+        let mut resp = RateLimitResponse::new();
+        resp.overall_code = RateLimitResponse_Code::OK;
+        resp.response_headers_to_add
+            .push(header("test", "some value"));
+        resp.response_headers_to_add
+            .push(header("other", "header value"));
+        let buffer = resp.write_to_bytes().unwrap();
+        // assert_eq!(b"", buffer.as_slice())
+    }
+
+    fn header(key: &str, value: &str) -> HeaderValue {
+        let mut header = HeaderValue::new();
+        header.key = key.to_string();
+        header.value = value.to_string();
+        header
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,7 +19,12 @@ mod tests {
         resp.response_headers_to_add
             .push(header("other", "header value"));
         let buffer = resp.write_to_bytes().unwrap();
-        // assert_eq!(b"", buffer.as_slice())
+        let expected: [u8; 45] = [
+            8, 1, 26, 18, 10, 4, 116, 101, 115, 116, 18, 10, 115, 111, 109, 101, 32, 118, 97, 108,
+            117, 101, 26, 21, 10, 5, 111, 116, 104, 101, 114, 18, 12, 104, 101, 97, 100, 101, 114,
+            32, 118, 97, 108, 117, 101,
+        ];
+        assert_eq!(expected, buffer.as_slice())
     }
 
     fn header(key: &str, value: &str) -> HeaderValue {
--- a/tests/rate_limited.rs
+++ b/tests/rate_limited.rs
@@ -276,8 +276,16 @@ fn it_passes_additional_headers() {
 
     module
         .call_proxy_on_response_headers(http_context, 0, false)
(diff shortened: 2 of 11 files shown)


tests/rate_limited.rs
```diff
@@ -0,0 +1,325 @@
...
+        )
+        .returning(Some(42))
+        .expect_log(
+            Some(LogLevel::Info),
+            Some("Initiated gRPC call (id# 42) to Limitador"),
+        )
+        .execute_and_expect(ReturnType::Action(Action::Pause))
+        .unwrap();
+
+    let grpc_response: [u8; 45] = [
```
eguzki, line 295: Why not encode like this? 

```rust
        let mut resp = RateLimitResponse::new();
        resp.overall_code = RateLimitResponse_Code::OK;
        resp.response_headers_to_add
            .push(header("test", "some value"));
        resp.response_headers_to_add
            .push(header("other", "header value"));
        let buffer = resp.write_to_bytes().unwrap();
        let grpc_response: [u8; 45] = buffer.as_slice();
```

I am not very familiar at reading bytes

tests/rate_limited.rs
```diff
@@ -0,0 +1,325 @@
+use proxy_wasm_test_framework::tester;
+use proxy_wasm_test_framework::types::{Action, BufferType, LogLevel, MapType, ReturnType};
+use serial_test::serial;
+use std::path::Path;
+
+fn wasm_module() -> String {
+    let wasm_file = Path::new("target/wasm32-unknown-unknown/release/wasm_shim.wasm");
```
eguzki, line 7: This means that the code tested is the actual (byte)code compiled with `wasm32-unknown-unknown` build target? 

Weird, but good enough for now I guess.

eguzki (commented):
Super cool!

great stuff! I say we go with this. 

minor questions dropped (so I can understand)

tests/rate_limited.rs
alexsnaps: Yes it is… Sorry if that was unclear, it's something I tried to explain early on when I first share this:

This embeds a runtime, [wasmtime](https://crates.io/crates/wasmtime), loads the wasm module and "pretends" to be a host, thru the [test-framework](https://github.com/Kuadrant/wasm-test-framework). It is complete blackbox testing, so we can't put the underlying code under test… we _could_ do assertion about the state of the memory in the module (the same was as the test framework reads & write to the wasm module's memory), but I think that'd be pushing it too far.

src/lib.rs
```diff
@@ -4,3 +4,33 @@ mod filter;
...
+
+    #[test]
+    fn grpc() {
+        let mut resp = RateLimitResponse::new();
+        resp.overall_code = RateLimitResponse_Code::OK;
+        resp.response_headers_to_add
+            .push(header("test", "some value"));
+        resp.response_headers_to_add
+            .push(header("other", "header value"));
+        let buffer = resp.write_to_bytes().unwrap();
```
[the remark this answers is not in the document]
alexsnaps: See common on [here](https://github.com/Kuadrant/wasm-shim/pull/36#discussion_r1234244329)

tests/rate_limited.rs
alexsnaps: We can't because of the lib type (`cdylib`), the tests in `tests` can't reference the code under test for… reasons. I couldn't get it to target another type, because of unresolved link time deps. That code, the envoy types, are part of the wasm-shim code and not a dependency, as they are generated using `protoc`.
As I mentioned in the sync, we could generate these twice, once as we do now, and another time in the `tests` dir, so we could use them there. But, just like for the PluginConfiguration, which uses a JSON format, I decided to use the serialized form directly, i.e. protobuf… hence the test in `lib.rs` that asserts that this is a valid byte sequence.

alexsnaps:
@eguzki Tried to address your questions in my responses, I hope this makes some sense. We can always hop on a call if that helps… I know this is somewhat surprising.

eguzki (approved):
This is already a good (and big) step forward. 

I have to admit that it feels that writing tests is gonna be very verbose having a lot of boilerplate code to write. 

Fortunately we can test the wasm module as a black box, using Rust and tests will be very quick. The framework allows adding expectations on hostcall methods as well as their input params and also configuring return values of those hostcalls.

PR #37, opened 2023-07-18
Wasm configuration v2

### LINKED ISSUE #35: WASM Configuration v2
### Current configuration struct

Example:
```yaml
failure_mode_deny: true
rate_limit_policies:
  - name: toystore
    hostnames: ["*.toystore.com"]
    gateway_actions:
      - rules:
          - paths: ["/admin/toy"]
            methods: ["GET"]
            hosts: ["pets.toystore.com"]
        configurations:
          - actions:
            - generic_key:
                descriptor_key: admin
                descriptor_value: "1"
```

### Proposal for a new configuration struct to support [RLP v2](https://github.com/Kuadrant/kuadrant-operator/issues/156)

**Motivation** 
[RFC for the RLP v2](https://github.com/Kuadrant/architecture/blob/main/rfcs/0001-rlp-v2.md) brings `routeSelectors` and `when` conditions to qualify traffic to be rate limited. Even thought it is technically possible to implement these rules on the Limitador side using Limitador's conditions, the wasm-shim has all the available info to implement those rules for qualifying traffic at the gateway, thus saving rate limiting traffic at the Limitador side. Current wasm-shim rules based on `paths`, `methods` and `hosts` are not enough to support `routeSelectors` and `when` conditions. 

Example to illustrate the new wasm shim configuration struct

```yaml
failureMode: deny
rateLimitPolicies:
  - name: rlp-ns-A/rlp-name-A
    domain: rlp-ns-A/rlp-name-A
    service: rate-limit-cluster
    hostnames: ["*.toystore.com"]
    rules:   
      - conditions:      
        - allOf:
          - selector: <value>
            operator: <value>
            value: <value>
          - selector: <value>
            operator: <value>
            value: <value>
        - allOf:
          - selector: <value>
            operator: <value>
            value: <value>
          - selector: <value>
            operator: <value>
            value: <value>
        data:
        - selector:
            selector: <value>       # Any attribute from a well-known specification TBD ([ref](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes#arch-overview-request-attributes))
            key: <value>             # Optional. If not set it defaults to `selector` field value as the descriptor key.
            default: <value>        # Optional. An optional value to use if metadata_key is empty. If not set and no value is present under the metadata_key then no descriptor is generated.
        - static:   
            key: <value>             # Required
            value: <value>          # Required
```

eguzki:
### What

Fixes #35 
Depends on https://github.com/Kuadrant/wasm-test-framework/pull/1 for e2e tests to pass

Example to illustrate the new wasm shim configuration struct

```yaml
failureMode: deny
rateLimitPolicies:
  - name: rlp-ns-A/rlp-name-A
    domain: rlp-ns-A/rlp-name-A
    service: rate-limit-cluster
    hostnames: ["*.toystore.com"]
    rules:   
      - conditions:      
        - allOf:
          - selector: <value>
            operator: <value>
            value: <value>
          - selector: <value>
            operator: <value>
            value: <value>
        - allOf:
          - selector: <value>
            operator: <value>
            value: <value>
          - selector: <value>
            operator: <value>
            value: <value>
        data:
        - selector:
            selector: <value>       # Any attribute from a well-known specification TBD ([ref](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes#arch-overview-request-attributes))
            key: <value>             # Optional. If not set it defaults to `selector` field value as the descriptor key.
            default: <value>        # Optional. An optional value to use if metadata_key is empty. If not set and no value is present under the metadata_key then no descriptor is generated.
        - static:   
            key: <value>             # Required
            value: <value>          # Required
```
* Upgrade `proxy-wasm` to 0.2.1. Hit this issue https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/154 and latest release includes the fix

TODO:
- [X] rate limit Wasm shim V2 new configuration
- [x] `matches` condition selector operator implementation
- [x] Tokenize `selector` fields with escape char `\`. Example: Input: `this.is.a.exam\.ple`  ->  Retuns: `["this", "is", "a", "exam.ple"]`. Some path segments include dot `.` char in them. For instance envoy filter names: `envoy.filters.http.header_to_metadata`
- [ ] ~Support data types other than `string` from envoy attributes https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes~ (opened issue https://github.com/Kuadrant/wasm-shim/issues/38 for that)
- [x] e2e tests
- [x] Doc


guicassolato:
LGTM so far, @eguzki. Nice job!

utils/docker-compose/envoy.yaml
```diff
@@ -50,138 +50,180 @@ static_resources:
...
                                     {
-                                      "actions": [
+                                      "allOf": [
                                         {
-                                          "generic_key": {
-                                            "descriptor_key": "some_generic_key",
-                                            "descriptor_value": "some_generic_value"
-                                          }
+                                          "selector": "request.url_path",
+                                          "operator": "starts_with",
```
guicassolato, line 84: RLPv2 [specifies](https://github.com/Kuadrant/kuadrant-operator/blob/529ae66e7514d5d28df76b1d039ac21136a1babf/api/v1beta2/ratelimitpolicy_types.go#L48) and effectively the Kuadrant Operator [generates](https://github.com/Kuadrant/kuadrant-operator/blob/529ae66e7514d5d28df76b1d039ac21136a1babf/pkg/rlptools/wasm/types.go#L15) for the WasmPlugin config `startswith`, not `starts_with`.

In fact, while testing this branch, I've observed the following error thrown by the gateway:

```
2023-07-25T12:20:20.200842Z     warning envoy wasm external/envoy/source/extensions/common/wasm/context.cc:1151 wasm log: failed to parse plugin config: unknown variant `startswith`, expected one of `eq`, `neq`, `starts_with`, `ends_with`, `matches` at line 1 column 184     thread=20
```

guicassolato (changes requested):
Operators `startswith` and `endswith` do not match with how they are generated by the controller.

src/configuration.rs
```diff
@@ -46,9 +46,9 @@ pub enum WhenConditionOperator {
     EqualOperator,
     #[serde(rename = "neq")]
     NotEqualOperator,
-    #[serde(rename = "starts_with")]
+    #[serde(rename = "startsWith")]
```
guicassolato, line 49:
```suggestion
    #[serde(rename = "startswith")]
```

😬

eguzki: already pushed the changes :)

eguzki: oh, you mean `startswith`, no camel case?

eguzki: ok ok... fixing it

eguzki:
@guicassolato ready for another review as you requested changes

eguzki:
green :partying_face:

src/filter/http_context.rs
```diff
@@ -63,227 +45,167 @@ impl Filter {
...
+        let attribute_path = tokenize_with_escaping(&p_e.selector, '.', '\\');
+        // convert a Vec<String> to Vec<&str>
+        // attribute_path.iter().map(AsRef::as_ref).collect()
+        match self.get_property(attribute_path.iter().map(AsRef::as_ref).collect()) {
+            None => {
+                debug!(
+                    "[context_id: {}]: selector not found: {}",
+                    self.context_id, p_e.selector
+                );
+                false
```
guicassolato, line 117: Not sure if we should default to "pattern expression does NOT apply" (and therefore nor the limit) when it fails to get the property. I see where this can be a problem such as in cases where the operator is `neq`. E.g.:

```yaml
- conditions:
  - allOf:
    - selector: missing.property
       operator: neq
       value: "x"
```

Technically, `missing.property != "x"`.

guicassolato (approved):

src/filter/http_context.rs
guicassolato: Thanks @alexsnaps!

PR #39, opened 2023-08-22
Fix return value when host property does not exist from data.selector

eguzki:
### What

Selectors in `rules[].data.selector[]` have a `default` named attribute to cover the use case where the property does not exist https://github.com/Kuadrant/wasm-shim/blob/main/src/configuration.rs#L7-L20

```rust
pub struct SelectorItem {
    // Selector of an attribute from the contextual properties provided by kuadrant
    // during request and connection processing
    pub selector: String,

    // If not set it defaults to `selector` field value as the descriptor key.
    #[serde(default)]
    pub key: Option<String>,

    // An optional value to use if the selector is not found in the context.
    // If not set and the selector is not found in the context, then no data is generated.
    #[serde(default)]
    pub default: Option<String>,
}
```

When the selector property does not exist, the behavior is ruled by the `default` field. Currently the implementation was enforcing a default value (empty string `""`) defeating the purpose of the `default` attribute.

### Verification Steps

* Run development env
```
make development
```

The wasm configuration applied for `*.a.com` is
```json
{
                              "name": "rlp-ns-A/rlp-name-A",
                              "domain": "rlp-ns-A/rlp-name-A",
                              "service": "limitador",
                              "hostnames": ["*.a.com"],
                              "rules": [
                                {
                                  "data": [
                                    {
                                      "selector": {
                                        "selector": "unknown.path"
                                      }
                                    }
                                  ]
                                }
                              ]
                            }
```

*  Run request   
```
curl -H "Host: test.a.com" http://127.0.0.1:18000/get
```
Data selector should not generate return any value. Thus, descriptor should be empty and rate limiting service should not be called. Logs should not show any log from the limitador service or GRPC client.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ regex = "1"
 radix_trie = "0.2.1"
 
 [dev-dependencies]
-proxy-wasm-test-framework = { git = "https://github.com/Kuadrant/wasm-test-framework.git", branch = "kuadrant" }
+proxy-wasm-test-framework = { git = "https://github.com/Kuadrant/wasm-test-framework.git", branch = "fix-get-property-when-none" }
 serial_test = "2.0.0"
 
 [build-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ regex = "1"
 radix_trie = "0.2.1"
 
 [dev-dependencies]
-proxy-wasm-test-framework = { git = "https://github.com/Kuadrant/wasm-test-framework.git", branch = "fix-get-property-when-none" }
+proxy-wasm-test-framework = { git = "https://github.com/Kuadrant/wasm-test-framework.git", branch = "kuadrant" }
 serial_test = "2.0.0"
 
 [build-dependencies]
--- a/src/filter/http_context.rs
+++ b/src/filter/http_context.rs
@@ -164,8 +164,7 @@ impl Filter {
                                 self.context_id, selector_item.selector
                             );
                             match &selector_item.default {
-                                // skipping the entire descriptor
-                                None => "".to_string(),
+                                None => return None, // skipping the entire descriptor
                                 Some(default_value) => default_value.clone(),
                             }
                         }
@@ -175,8 +174,7 @@ impl Filter {
                             Err(e) => {
                                 debug!(
                                 "[context_id: {}]: failed to parse selector value: {}, error: {}",
-                                self.context_id, selector_item.selector, e
-                            );
+                                self.context_id, selector_item.selector, e);
                                 return None;
                             }
                             Ok(attribute_value) => attribute_value,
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -26,7 +26,7 @@ regex = "1"
 radix_trie = "0.2.1"
 
 [dev-dependencies]
-proxy-wasm-test-framework = { git = "https://github.com/Kuadrant/wasm-test-framework.git", branch = "kuadrant" }
+proxy-wasm-test-framework = { git = "https://github.com/Kuadrant/wasm-test-framework.git", branch = "fix-get-property-when-none" }
```
eguzki, line 29: this change will be reverted, only used for testing https://github.com/Kuadrant/wasm-test-framework/pull/2

src/filter/http_context.rs
```diff
@@ -164,8 +164,7 @@ impl Filter {
                                 self.context_id, selector_item.selector
                             );
                             match &selector_item.default {
-                                // skipping the entire descriptor
-                                None => "".to_string(),
+                                None => return None, // skipping the entire descriptor
```
guicassolato, line 167: I think returning an empty string was [intentional](https://github.com/Kuadrant/wasm-shim/commit/59033b0b4f89290391c8eee9acf00fde51fe6654), as part of avoiding the unsafe default of skipping the limit when a descriptor expected to exist is missing. Rather, it seemed safer to occasionally over-limit (i.e. by expanding the scope of a counter to a common, "catch-all", counter when a selector is evaluated to `""`, because the descriptor is missing), than to under-limit (by not enforcing any limit at all).

Either way, the user will have to debug and figure out why it's not rate limiting as expected.

eguzki: For selectors in conditions, it makes sense to return something when the property is not found in order to apply operators. However, for `data.selectors`, there is the `default` attribute to cover the use case where the property does not exist https://github.com/Kuadrant/wasm-shim/blob/main/src/configuration.rs#L7-L20

```rust
pub struct SelectorItem {
    // Selector of an attribute from the contextual properties provided by kuadrant
    // during request and connection processing
    pub selector: String,

    // If not set it defaults to `selector` field value as the descriptor key.
    #[serde(default)]
    pub key: Option<String>,

    // An optional value to use if the selector is not found in the context.
    // If not set and the selector is not found in the context, then no data is generated.
    #[serde(default)]
    pub default: Option<String>,
}
```

Check the doc string of the `default` field. 

IMO it makes sense to assume that if the property does not exist AND there is no default, no descriptor entry is generated. If you want to generate descriptor entry no matter the property exists or not, set `default`. 

At the kuadrant control plane level, when there is RLP CR counter, the `selector.data[].selector` item will be generated. If we want to reach limitador no matter the property exists or not, the controller should add a default. But that's control plane level call. IMO, the tool (the wasm shim) should cover all use cases.

guicassolato: Gotcha. Thanks for explaining.

eguzki:
requires https://github.com/Kuadrant/wasm-test-framework/pull/2

eguzki:
ready for review @alexsnaps

alexsnaps (approved):
So sorry @eguzki - I had like 3 tabs open with this PR 🤦

PR #41, opened 2023-08-25
Typing

alexsnaps:
Still work in progress, if only because of two things:

 - [x] Just as on `main` this returns an empty string on misses
 - [ ] The `as_literal` syntax will need to be supported on the limitador end too
 - [ ] Still working on adding support for the more complex types which are afaict protobuff encoded

### DIFF
--- a/src/envoy/properties.rs
+++ b/src/envoy/properties.rs
@@ -9,8 +9,7 @@ pub struct EnvoyTypeMapper {
 
 impl EnvoyTypeMapper {
     pub fn new() -> Self {
-        let mut properties: BTreeMap<String, Box<MapperFn>> =
-            BTreeMap::new();
+        let mut properties: BTreeMap<String, Box<MapperFn>> = BTreeMap::new();
         properties.insert("foo.bar".to_string(), Box::new(TypedProperty::string));
         properties.insert("foo.car".to_string(), Box::new(TypedProperty::integer));
         Self {
@@ -25,4 +24,3 @@ impl EnvoyTypeMapper {
         }
     }
 }
-
--- a/src/envoy/properties.rs
+++ b/src/envoy/properties.rs
@@ -10,8 +10,183 @@ pub struct EnvoyTypeMapper {
 impl EnvoyTypeMapper {
     pub fn new() -> Self {
         let mut properties: BTreeMap<String, Box<MapperFn>> = BTreeMap::new();
-        properties.insert("foo.bar".to_string(), Box::new(TypedProperty::string));
-        properties.insert("foo.car".to_string(), Box::new(TypedProperty::integer));
+        properties.insert(
+            "request.time".to_string(),
+            Box::new(TypedProperty::timestamp),
+        );
+
+        properties.insert("request.id".to_string(), Box::new(TypedProperty::string));
+        properties.insert(
+            "request.protocol".to_string(),
+            Box::new(TypedProperty::string),
+        );
+        properties.insert(
+            "request.scheme".to_string(),
+            Box::new(TypedProperty::string),
+        );
+        properties.insert("request.host".to_string(), Box::new(TypedProperty::string));
+        properties.insert(
+            "request.method".to_string(),
+            Box::new(TypedProperty::string),
+        );
+        properties.insert("request.path".to_string(), Box::new(TypedProperty::string));
+        properties.insert(
+            "request.url_path".to_string(),
+            Box::new(TypedProperty::string),
+        );
+        properties.insert("request.query".to_string(), Box::new(TypedProperty::string));
+        properties.insert(
+            "request.referer".to_string(),
+            Box::new(TypedProperty::string),
+        );
+        properties.insert(
+            "request.useragent".to_string(),
+            Box::new(TypedProperty::string),
+        );
+        properties.insert("request.body".to_string(), Box::new(TypedProperty::string));
+        properties.insert(
+            "source.address".to_string(),
+            Box::new(TypedProperty::string),
+        );
+        properties.insert(
+            "source.service".to_string(),
+            Box::new(TypedProperty::string),
+        );
+        properties.insert(
+            "source.principal".to_string(),
+            Box::new(TypedProperty::string),
+        );
+        properties.insert(
+            "source.certificate".to_string(),
+            Box::new(TypedProperty::string),
+        );
+        properties.insert(
+            "destination.address".to_string(),
+            Box::new(TypedProperty::string),
+        );
+        properties.insert(
+            "destination.service".to_string(),
+            Box::new(TypedProperty::string),
+        );
+        properties.insert(
+            "destination.principal".to_string(),
+            Box::new(TypedProperty::string),
+        );
+        properties.insert(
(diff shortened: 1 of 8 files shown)


src/filter/http_context.rs
```diff
@@ -152,35 +130,11 @@ impl Filter {
...
-                        .get_property(attribute_path.iter().map(AsRef::as_ref).collect())
-                    {
-                        None => {
-                            debug!(
-                                "[context_id: {}]: selector not found: {}",
-                                self.context_id, selector_item.selector
-                            );
-                            match &selector_item.default {
-                                // skipping the entire descriptor
-                                None => "".to_string(),
```
alexsnaps, line 168: That's a lie... but it should skip it yes?

src/filter/http_context.rs
```diff
@@ -195,6 +149,23 @@ impl Filter {
         Some(res)
     }
 
+    fn get_typed_property(&self, path: &str) -> TypedProperty {
```
alexsnaps, line 152: This should then return an `Option<TypedProperty>` and act accordingly on the call site, wdyt?

src/filter/http_context.rs
```diff
@@ -195,6 +149,23 @@ impl Filter {
...
+        let tokens = tokenize_with_escaping(path, '.', '\\');
+        match self.get_property(tokens.iter().map(AsRef::as_ref).collect()) {
+            None => {
+                debug!(
+                    "[context_id: {}]: selector not found: {path}, defaulting to ``",
+                    self.context_id
+                );
+                TypedProperty::String("".to_string())
```
alexsnaps, line 160: Ported from `main` but should return `None` I think...

src/filter/http_context.rs
```diff
@@ -195,6 +149,23 @@ impl Filter {
...
+            }
+            Some(attribute_bytes) => match self.property_mapper.typed(path, attribute_bytes) {
+                Ok(tp) => tp,
+                Err(raw) => TypedProperty::string(raw),
```
alexsnaps, line 164: So, if we fail to map a known property, we try to make it a `String` which in turn may fail, when there is an invalid byte sequence, which then becomes a `Bytes`. I added support for that in the `as_literal` tho unsure what one would do this that...

src/envoy/properties.rs
```diff
@@ -0,0 +1,201 @@
...
+            Box::new(TypedProperty::complex_map),
+        );
+        properties.insert(
+            "auth.response".to_string(),
+            Box::new(TypedProperty::complex_map),
+        );
+        properties.insert(
+            "auth.callbacks".to_string(),
+            Box::new(TypedProperty::complex_map),
+        );
```
guicassolato, line 177: `auth.*` are not properties that can be requested directly to the host through the ABI, right? For that reason they were unchecked for "RL" in [RFC 0002](https://github.com/Kuadrant/architecture/blob/main/rfcs/0002-well-known-attributes.md#auth-attributes) for now. Seeing them here makes me think we may want to progress to the point where these attributes are abstractions that translate to `metadata.filter_metadata.envoy\.filters\.http\.ext_authz.*` perhaps?

Do we need a special `TypedProperty` function to handle them maybe?

Another option would be the "translation" to happen in the policy controller, and therefore the wasm shim won't ever even need to know about them.

src/envoy/properties.rs
```diff
@@ -0,0 +1,201 @@
...
+        properties.insert(
+            "connection.mtls".to_string(),
+            Box::new(TypedProperty::boolean),
+        );
+
+        properties.insert(
+            "request.raw_body".to_string(),
+            Box::new(TypedProperty::bytes),
+        );
+        properties.insert("auth.identity".to_string(), Box::new(TypedProperty::bytes));
```
guicassolato, line 188: https://github.com/Kuadrant/wasm-shim/pull/41/files#r1307418336

src/filter/http_context.rs
eguzki: Then maybe return `Result<Option<TypedProperty>, Vec<u8>>` instead of `option` and leave the decision to the caller? 

In any case, what I would do is logging error saying.. "property A could not be parsed". 

When the caller is the pattern expression, the error could be "overlooked" as `""`, while the descriptor builder can fallback to `default` field value.

src/typing.rs
```diff
@@ -0,0 +1,124 @@
...
+        if bytes.len() == 8 {
+            TypedProperty::Integer(i64::from_le_bytes(
+                bytes[..8].try_into().expect("This has to be 8 bytes long!"),
+            ))
+        } else {
+            TypedProperty::bytes(bytes)
+        }
+    }
+
+    pub fn timestamp(bytes: Vec<u8>) -> Self {
```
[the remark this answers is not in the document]
eguzki: Additionally, isn't this Envoy Specific implementation? Do we want to encapsulate envoy specific things?

eguzki (commented):
I think that at the `http_context.rs` level:
* pattern expressions are evaluated with gateway specific implementation returning true or false. It would also require a gateway specific implementation for the parsing of the `value` field (that could also defined as byte array to be whatever, ie. json int, string, array or object).
* on descriptor building, data selectors call gateway specific implementation returning `Option<String>` (as RLS only accepts strings). (Or maybe `Result<Option<String>, <Vec<u8>>` for dealing with parsing errors)

Needless to say the only gateway implementation for now is envoy.

Does it makes sense what I say :question:

src/filter/http_context.rs
alexsnaps: I prefer the current approach which clearly separate the two concerns... 
Now I could _not_ fold the error case in the `TypedProperty::string` function, but I am of the opinion that's the behavior we should aim for: unknown is tried as a `String`, if it isn't proper UTF-8, we fall back to string encoded bytes (see `TypedProperty::as_string` or `::as_literal` for `TypedProperty::Bytes`)

src/envoy/properties.rs
alexsnaps: This actually merely defines the typing... but you are right that I conflated the typing with the host. `Filter::get_typed_property` is actually what decides where the property is to be sourced. 

Maybe I should treat all auth related one differently... tho they'd still use some typed mapper function just as the envoy properties do... 🤔

src/filter/http_context.rs
```diff
@@ -152,33 +133,14 @@ impl Filter {
...
-                                debug!(
-                                "[context_id: {}]: failed to parse selector value: {}, error: {}",
-                                self.context_id, selector_item.selector, e);
-                                return None;
-                            }
-                            Ok(attribute_value) => attribute_value,
-                        },
+                    let typed_property = match self.get_typed_property(&selector_item.selector) {
+                        Some(typed_property) => typed_property,
+                        None => return None,
```
eguzki, line 138: where is `&selector_item.default` being used when the typed property is not found?

alexsnaps: Not there right now, as it itself introduces a rather non neglectable refactoring of this PR: tl;dr it needs type inference and coercing from the now not `String` value, but a Value literal.

alexsnaps:
Superseded by #60

PR #49, opened 2024-03-19
Propagate opentelemetry tracing headers in requests to limitador

adam-cattermole:


src/filter/http_context.rs
```diff
@@ -207,6 +214,15 @@ impl HttpContext for Filter {
     fn on_http_request_headers(&mut self, _: usize, _: bool) -> Action {
         info!("on_http_request_headers #{}", self.context_id);
 
+        let req_headers = self.get_http_request_headers();
+        for (header, value) in req_headers.iter() {
+            match header.as_str() {
+                "traceparent" | "tracestate" | "baggage" => {
```
eguzki, line 220: `self.get_http_request_headers` returns lowercase header names? in other words, if the header is `TraceParent`, is the value captured? According to https://w3c.github.io/trace-context/#traceparent-header

```
The header name is [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive). That is, TRACEPARENT, TraceParent, and traceparent are considered the same header. The header name is a single word; it does not contain any delimiters such as a hyphen.
```

adam-cattermole: Yep from what I could tell no matter the provided format the headers were retrieved in lowercase - perhaps due to this https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/header_casing

I could add a `.to_lowercase()` before the match for safety in case though. Also considering moving the three header strings out to constants

eguzki: if it's only used there, no need for a constant.

eguzki:
### verification steps

* Run devel env

```
make development
```
> This is currently failing but easily fixed with
```diff
diff --git a/Makefile b/Makefile
index 7808041..5362a26 100644
--- a/Makefile
+++ b/Makefile
@@ -61,10 +61,10 @@ $(WASM_RELEASE_PATH): $(RUST_SOURCES)
        make -C $(PROJECT_PATH) -f $(MKFILE_PATH) build
 
 development: $(WASM_RELEASE_PATH)
-       docker-compose up
+       docker compose up
 
 stop-development:
-       docker-compose down
+       docker compose down
 
 # get-protoc will download zip from $2 and install it to $1.
 define get-protoc
```
* Send a request that will hit limitador with a `Traceparent` header (upper case `T` used on purpose)
```
curl -H "Host: test.c.com" -H "x-forwarded-for: 127.0.0.1" -H "My-Custom-Header-01: my-custom-header-value-01" -H 'Traceparent: 00-80e1afed08e019fc1110464cfa66635c-7a085853722dc6d2-01' -H "x-dyn-user-id: bob" http://127.0.0.1:18000/get
```

Expected to see in limitador's logs the `traceparent`

```
limitador-1  | 2024-03-21T16:28:47.739377Z DEBUG should_rate_limit: limitador_server::envoy_rls::server: Request received: Request { metadata: MetadataMap { headers: {"te": "trailers", "grpc-timeout": "5000m", "content-type": "application/grpc", "traceparent": "00-80e1afed08e019fc1110464cfa66635c-7a085853722dc6d2-01", "x-envoy-internal": "true", "x-forwarded-for": "172.21.0.4", "x-envoy-expected-rq-timeout-ms": "5000"} }, message: RateLimitRequest { domain: "rlp-ns-C/rlp-name-C", descriptors: [RateLimitDescriptor { entries: [Entry { key: "limit_to_be_activated", value: "1" }], limit: None }, RateLimitDescriptor { entries: [Entry { key: "source.address", value: "127.0.0.1:0" }], limit: None }, RateLimitDescriptor { entries: [Entry { key: "request.headers.My-Custom-Header-01", value: "my-custom-header-value-01" }], limit: None }, RateLimitDescriptor { entries: [Entry { key: "user_id", value: "bob" }], limit: None }], hits_addend: 1 }, extensions: Extensions }

```

eguzki (approved):

eguzki:
@adam-cattermole nitpick: does it makes sense to use some external library, as a new dep [strum](https://docs.rs/strum/0.26.2/strum/), for such a simple thing as to iterate few values?

adam-cattermole:
> @adam-cattermole nitpick: does it makes sense to use some external library, as a new dep [strum](https://docs.rs/strum/0.26.2/strum/), for such a simple thing as to iterate few values?

I had the same thought on this - I'm wondering if there's a nice alternative that would allow us to simply add/remove from the `TracingHeader` enum without changing anything else.

src/filter/http_context.rs
```diff
@@ -6,33 +6,36 @@ use crate::envoy::{
...
 use std::time::Duration;
-use strum::IntoEnumIterator;
-use strum_macros::EnumIter;
 
 const RATELIMIT_SERVICE_NAME: &str = "envoy.service.ratelimit.v3.RateLimitService";
 const RATELIMIT_METHOD_NAME: &str = "ShouldRateLimit";
 
 // tracing headers
-#[derive(EnumIter)]
+#[derive(Clone)]
```
alexsnaps, line 22: Not hurtful, but I don't know we clone that anywhere...

alexsnaps (approved):

PR #52, opened 2024-04-29
Wasm server

eguzki:
All the info available https://github.com/Kuadrant/kuadrant-operator/pull/593


Dockerfile.server
```diff
@@ -0,0 +1,37 @@
+# ------------------------------------------------------------------------------
```
eguzki, line 1: I would like to reuse the building stage, which has been copied&pasted from the `Dockerfile`. Ideas?

eguzki:
decided not to build custom service to serve wasm binaries

PR #54, opened 2024-05-29
Fix tests

eguzki:
### What

Fix broken tests since https://github.com/Kuadrant/wasm-shim/pull/49. That included more host calls that are not expected by the test framework.

Depends on https://github.com/Kuadrant/wasm-test-framework/pull/3 (test will fail until PR is merged in the wasm-test-framework)

Additionally, create it's own GH actions workflow for tests. The existing workflow was running nightly basis and it was automatically disabled after 60 days of inactivity on the project.


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,33 @@
+---
+name: Tests
+
+on:
+  push:
+    branches:
+      - '*'
```
didierofrivia, line 7:
```suggestion
      - 'main'
```

didierofrivia (approved):
Maybe change the workflow event on push only to main branches and it´s good to go!

PR #57, opened 2024-06-14
logging enhancements

eguzki:
### What

- http context logging at `debug` level. 
- http context id consistent for all loggers with the format `#{}` at the end of the log line.
- Fix clippy complaining about rate limit policy name not being used. Added the name to a log line (at debug level).

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -31,7 +31,7 @@ extern "C" fn start() {
         proxy_wasm::hostcalls::log(LogLevel::Critical, &panic_info.to_string()).unwrap();
     }));
     proxy_wasm::set_root_context(|context_id| -> Box<dyn RootContext> {
-        info!("set_root_context #{}", context_id);
+        info!("#{} set_root_context", context_id);
         Box::new(FilterRoot {
             context_id,
             config: Rc::new(FilterConfig::new()),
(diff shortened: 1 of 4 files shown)


eguzki:
TODO: review logging traces again under heavy traffic stress. When multiple requests hit "at the same time", try to see if one single request can be traced for debugging.

adam-cattermole (approved):
Changes look good to me. I guess my only question is the reason for moving the context id to the end of the line? I guess it's personal preference? I'm undecided whether it's easier to identify the logs you're looking for with the identifier at the start vs at the end

adam-cattermole:
I guess to add to my thoughts there - if there is going to be logs with different context id's interleaved would it not be easier to debug with the context at the start instead of having to look at the end of each log first?

eguzki:
> ere is going to be logs with different context

Fair point. Not a strong opinion here. I saw there was  some logs doing that and I tried to apply some consistency. 

I will add at the start

eguzki:
For the record: the context id is related to the connection. Multiple (slow) requests share the same context id (tested).

Following same approach as the example from rust-sdk lib https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/main/examples/http_headers/src/lib.rs#L46-L48

```rust
for (name, value) in &self.get_http_request_headers() {
            info!("#{} -> {}: {}", self.context_id, name, value);
}
```

adam-cattermole (approved):

didierofrivia (approved):
🔎

PR #60, opened 2024-07-10
Typing cel

### LINKED ISSUE #38: Selectors: support data types other than string from envoy attributes
Currently, only selectors that return string data type can be used.

From the [Envoys attributes](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes), examples can be `request.path`, `source.address` or `request.headers.<header-name>`. 

However, other attributes return a different type of data. Instances can be `request.time`, `request.size`, `source.port` or metadata based attributes.

alexsnaps:
fixes #38


src/configuration.rs
```diff
@@ -54,34 +132,193 @@ pub enum WhenConditionOperator {
...
+        "ratelimit.hits_addend" => ValueType::Int,
+        "request.headers" => ValueType::Map,
+        "request.context_extensions" => ValueType::Map,
+        "source.labels" => ValueType::Map,
+        "destination.labels" => ValueType::Map,
+        "filter_state" => ValueType::Map,
+        "connection.mtls" => ValueType::Bool,
+        "request.raw_body" => ValueType::Bytes,
+        "auth.identity" => ValueType::Bytes,
+        _ => ValueType::Bytes,
```
alexsnaps, line 320: This probably is best to first try a `String` and fallback to `Bytes` if not a valid utf-8 sequence... 
That's mean that someone accessing a value "deeper" within a well-know attribute (e.g. `metadata.filter_metadata.envoy\.filters\.http\.ext_authz.identity.userid`), it try to parse it as a string first

Cargo.toml
```diff
@@ -31,6 +31,9 @@ thiserror = "1.0"
 regex = "1"
 radix_trie = "0.2.1"
 const_format = "0.2.31"
+chrono = { version = "0.4.38", default-features = false, features = ["alloc", "std"] }
+cel-interpreter = { git = "https://github.com/clarkmcc/cel-rust", branch = "master" }
+cel-parser = { git = "https://github.com/clarkmcc/cel-rust.git", branch = "master" }
```
alexsnaps, line 36: We need something newer than `v0.8.0`, as we need this [fix](https://github.com/clarkmcc/cel-rust/issues/62)

src/configuration.rs
```diff
@@ -561,4 +966,210 @@ mod test {
...
+            assert_eq!(
+                p.eval((true as u8).to_le_bytes().to_vec()),
+                Ok(true),
+                "Expression: {:?}",
+                p.compiled.get()
+            )
+        }
+
+        #[test]
+        #[ignore]
```
alexsnaps, line 1152: I think this might require another [fix](https://github.com/clarkmcc/cel-rust/issues/67) but not a huge deal, as I don't think that's really useful right now anyways…

Cargo.toml
didierofrivia: Last week was released [v0.8.1](https://github.com/clarkmcc/cel-rust/releases/tag/v0.8.1) which includes a bunch of your fixes :)

alexsnaps:
I'll rebase this work and fix the dep.

didierofrivia (approved):
📝

PR #64, opened 2024-08-07
Refactoring for external auth. Vol I

didierofrivia:
This PR is the first of hopefully not too many that aims to re-arrange and open the code in order to implement External Auth.

Notes:
* Renames `RateLimitPolicy` struct to `Policy`
* Keeps config with `rateLimitPolicies` field, but renames it to  `policies` field within the struct `PluginConfiguration` when deserializing.
* Delegates the building of descriptors to the `Policy` instead of `Filter`
* Missing configuration change, TODO alongside a PR to Kuadrant Operator. Its still functional since it's serde::renaming the field

### DIFF
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -485,7 +485,7 @@ pub struct Rule {
 
 #[derive(Deserialize, Debug, Clone)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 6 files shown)


adam-cattermole (approved):
lgtm

PR #65, opened 2024-08-09
Move attribute parsing to attribute trait

adam-cattermole:
(builds on top of #64 refactoring-1-bis)

For the auth implementation I wanted to be able to retrieve/parse the different attributes from `filter.get_property` without need to evaluate a `PatternExpression` so I decided to move this to it's own trait with an implementation for each type (where the implementation comes from the original one within the pattern expression).

This then introduces a generic `get_attribute` function which can be used like this

```
get_attribute(filter, "request.host")
```

To return a `Result<String,String>` in this case with the property.

Looking for all feedback as I'm unsure if this is the correct approach to do this.

### DIFF
--- a/src/attribute.rs
+++ b/src/attribute.rs
@@ -104,6 +104,7 @@ impl Attribute for DateTime<FixedOffset> {
     }
 }
 
+#[allow(dead_code)]
 pub fn get_attribute<T>(f: &Filter, attr: &str) -> Result<T, String>
 where
     T: Attribute,
--- a/src/attribute.rs
+++ b/src/attribute.rs
@@ -24,7 +24,7 @@ impl Attribute for i64 {
     fn parse(raw_attribute: Vec<u8>) -> Result<Self, String> {
         if raw_attribute.len() != 8 {
             return Err(format!(
-                "Int value expected to be 8 bytes, but got {}",
+                "parse: Int value expected to be 8 bytes, but got {}",
                 raw_attribute.len()
             ));
         }
@@ -40,7 +40,7 @@ impl Attribute for u64 {
     fn parse(raw_attribute: Vec<u8>) -> Result<Self, String> {
         if raw_attribute.len() != 8 {
             return Err(format!(
-                "UInt value expected to be 8 bytes, but got {}",
+                "parse: UInt value expected to be 8 bytes, but got {}",
                 raw_attribute.len()
             ));
         }
@@ -56,7 +56,7 @@ impl Attribute for f64 {
     fn parse(raw_attribute: Vec<u8>) -> Result<Self, String> {
         if raw_attribute.len() != 8 {
             return Err(format!(
-                "Float value expected to be 8 bytes, but got {}",
+                "parse: Float value expected to be 8 bytes, but got {}",
                 raw_attribute.len()
             ));
         }
@@ -78,7 +78,7 @@ impl Attribute for bool {
     fn parse(raw_attribute: Vec<u8>) -> Result<Self, String> {
         if raw_attribute.len() != 1 {
             return Err(format!(
-                "Bool value expected to be 1 byte, but got {}",
+                "parse: Bool value expected to be 1 byte, but got {}",
                 raw_attribute.len()
             ));
         }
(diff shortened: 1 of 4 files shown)


src/policy.rs
```diff
@@ -125,31 +126,7 @@ impl Policy {
...
-                    let value = match filter.get_property(attribute_path.tokens()) {
-                        None => {
-                            debug!(
-                                "#{} build_single_descriptor: selector not found: {}",
-                                filter.context_id, attribute_path
-                            );
-                            match &selector_item.default {
-                                None => return None, // skipping the entire descriptor
-                                Some(default_value) => default_value.clone(),
-                            }
```
adam-cattermole, line 138: Woops this should be fixed

src/policy.rs
```diff
@@ -139,16 +140,7 @@ impl Policy {
...
-                            Err(e) => {
-                                debug!(
-                                    "#{} build_single_descriptor: failed to parse selector value: {}, error: {}",
-                                    filter.context_id, attribute_path, e
-                                );
-                                return None;
-                            }
-                            Ok(attribute_value) => attribute_value,
-                        },
+                        Some(attribute_bytes) => Attribute::parse(attribute_bytes).ok()?,
```
adam-cattermole, line 143: I'm not 100% here, because this could take advantage of the `type_of` to determine the type instead of assuming it's a `String`

src/attribute.rs
```diff
@@ -0,0 +1,119 @@
...
+            raw_attribute.as_slice()[..8]
+                .try_into()
+                .expect("This has to be 8 bytes long!"),
+        );
+        Ok(DateTime::from_timestamp_nanos(nanos).into())
+    }
+}
+
+#[allow(dead_code)]
+pub fn get_attribute<T>(f: &Filter, attr: &str) -> Result<T, String>
```
alexsnaps, line 108: This is apparently still `dead_code`, but I also wonder how this is going to be useful... `<T>` needs to be known at compile time here, so that we _need_ to know at method invocation time what the `type` of `attr`... is this ever going to be the case that this is known _before_ runtime?

src/configuration.rs
```diff
@@ -168,47 +170,13 @@ impl PatternExpression {
...
-                    Value::Timestamp(time)
-                }
-            }
+            ValueType::String => Value::String(Arc::new(Attribute::parse(raw_attribute)?)),
+            ValueType::Int => Value::Int(Attribute::parse(raw_attribute)?),
+            ValueType::UInt => Value::UInt(Attribute::parse(raw_attribute)?),
+            ValueType::Float => Value::Float(Attribute::parse(raw_attribute)?),
+            ValueType::Bytes => Value::Bytes(Arc::new(Attribute::parse(raw_attribute)?)),
+            ValueType::Bool => Value::Bool(Attribute::parse(raw_attribute)?),
+            ValueType::Timestamp => Value::Timestamp(Attribute::parse(raw_attribute)?),
```
alexsnaps, line 179: Nice ❤️

alexsnaps (approved):

src/attribute.rs
```diff
@@ -0,0 +1,119 @@
...
+use chrono::{DateTime, FixedOffset};
+use proxy_wasm::traits::Context;
+
+pub trait Attribute {
+    fn parse(raw_attribute: Vec<u8>) -> Result<Self, String>
+    where
+        Self: Sized;
+}
+
+impl Attribute for String {
```
alexsnaps, line 12: I think this could "just" be the `TryInto` trait, but it is that against the [orphan rules](https://doc.rust-lang.org/reference/items/implementations.html?highlight=orphan#orphan-rules)

src/attribute.rs
adam-cattermole: As an example when I was building some response stuff for auth I was using it like this (providing a default to infer types):

```
attr.set_source(build_peer(
    get_attribute(&self, "source.address").unwrap_or("".to_string()),
    get_attribute(&self, "source.port").unwrap_or(0i64) as u32,
));
```

But then again now I think about it, these should probably remain unset if we can't get the attribute and not be a default value.

alexsnaps: Yeah, that actually makes sense... forgot that in the case of auth, there is a set of attributes (both their name _and_ type) that need to be retrieved no matter what... so yes, that `fn` would be useful for that. The `unwrap_or` I'm less confident about it indeed as well

didierofrivia (approved):
🍬

Issue #58, opened 2024-07-01
Ext_authz


guicassolato:
Make possible for the wasm-shim to issue gRPC Envoy [ext_authz](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/external_auth.proto#envoy-v3-api-msg-service-auth-v3-checkrequest) requests based on configuration.

Other than the main attributes `source`, `destination`, and `request` of the `CheckRequest` message, supplying values in the [`context_extensions`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/attribute_context.proto#envoy-v3-api-msg-service-auth-v3-attributecontext) field must as well be supported. This will be useful to hint the [authorization service lookup](https://docs.kuadrant.io/0.8.0/authorino/docs/architecture/#host-lookup) without the constraint on the host name as the only unique identifier of effective policies (AuthConfig CRs.)

david-martin:
@guicassolato Will this change mean that tracing behaves the same as with limitador?
That is, trace ids [do not propagate](https://github.com/envoyproxy/envoy/issues/22028) to wasm modules in Istio/Envoy, affecting trace continuity in authorino?
(in which case, we should update the tracing guide at https://docs.kuadrant.io/0.8.0/kuadrant-operator/doc/observability/tracing/ to mention this)

guicassolato:
> @guicassolato Will this change mean that tracing behaves the same as with limitador? That is, trace ids [do not propagate](https://github.com/envoyproxy/envoy/issues/22028) to wasm modules in Istio/Envoy, affecting trace continuity in authorino? (in which case, we should update the tracing guide at https://docs.kuadrant.io/0.8.0/kuadrant-operator/doc/observability/tracing/ to mention this)

It probably does, @david-martin. Thanks for pointing this out!

eguzki:
@didierofrivia I guess we can close this after https://github.com/Kuadrant/wasm-shim/pull/92. The RFE is

> Make possible for the wasm-shim to issue gRPC Envoy [ext_authz](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/external_auth.proto#envoy-v3-api-msg-service-auth-v3-checkrequest) requests based on configuration.

The wasm-shim supports that today with the configuration action
```yaml
"extensions": {
            "authorino": {
                "type": "auth",
                "endpoint": "authorino-cluster",
                "failureMode": "deny",
[...]

PR #66, opened 2024-08-12
Refactoring for external auth. Vol II

didierofrivia:
This PR follows the work done in https://github.com/Kuadrant/wasm-shim/pull/64, it aims to extract the grpc service logic.

With this refactor we now can build different implementations of `Service` in order to call external GRPC services to validate messages.

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -1,0 +1,8 @@
+pub(crate) mod rate_limit;
+
+use protobuf::Message;
+use proxy_wasm::types::Status;
+
+pub trait Service<M: Message> {
+    fn send(&self, message: M) -> Result<u32, Status>;
+}
--- a/src/service/rate_limit.rs
+++ b/src/service/rate_limit.rs
@@ -1,0 +1,105 @@
+use crate::envoy::{RateLimitDescriptor, RateLimitRequest};
+use crate::service::Service;
+use protobuf::{Message, RepeatedField};
+use proxy_wasm::hostcalls::dispatch_grpc_call;
+use proxy_wasm::types::Status;
+use std::time::Duration;
+
+const RATELIMIT_SERVICE_NAME: &str = "envoy.service.ratelimit.v3.RateLimitService";
+const RATELIMIT_METHOD_NAME: &str = "ShouldRateLimit";
+pub struct RateLimitService<'a> {
+    endpoint: String,
+    metadata: Vec<(&'a str, &'a [u8])>,
+}
+
+impl<'a> RateLimitService<'a> {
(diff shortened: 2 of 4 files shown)


src/service.rs
```diff
@@ -0,0 +1,22 @@
...
+    initial_metadata: Vec<(&str, &[u8])>,
+    message: Option<&[u8]>,
+    timeout: Duration,
+) -> Result<u32, Status>;
+
+pub trait Service {
+    fn call(
+        &self,
+        dispatcher: GrpcDispatcher,
+        message: Vec<u8>,
```
alexsnaps, line 19: This should probably of type `<M>`, that's on a per `Service<M>`

src/service/rate_limit.rs
```diff
@@ -0,0 +1,105 @@
...
+    upstream_name: &str,
+    initial_metadata: Vec<(&str, &[u8])>,
+    message: RateLimitRequest,
+) -> Result<u32, Status> {
+    let msg = Message::write_to_bytes(&message).unwrap(); // TODO(didierofrivia): Error Handling
+    dispatch_grpc_call(
+        upstream_name,
+        RATELIMIT_SERVICE_NAME,
+        RATELIMIT_METHOD_NAME,
+        initial_metadata,
```
alexsnaps, line 46: I guess this should be `self.metadata` and `initial_metadata` disappears from that method signature, right?

src/service/rate_limit.rs
```diff
@@ -0,0 +1,105 @@
+use crate::envoy::{RateLimitDescriptor, RateLimitRequest};
+use crate::service::Service;
+use protobuf::{Message, RepeatedField};
+use proxy_wasm::hostcalls::dispatch_grpc_call;
+use proxy_wasm::types::Status;
+use std::time::Duration;
+
+const RATELIMIT_SERVICE_NAME: &str = "envoy.service.ratelimit.v3.RateLimitService";
+const RATELIMIT_METHOD_NAME: &str = "ShouldRateLimit";
+pub struct RateLimitService<'a> {
```
alexsnaps, line 10: I think, if `metadata` is the "headers", and in our case the tracing ones, you can just claim ownership of these `str` and get rid of `<'a>`

src/service/rate_limit.rs
didierofrivia: I've made the `grpc_call` outside the `impl` block so we can mock it in the tests, that's why it doesn't take a `self` param. Should it be include it within the `RateLimitService` impl?

alexsnaps (approved):

PR #67, opened 2024-08-15
Integration development environment

adam-cattermole:
Added new make commands to:

* Deploy kind cluster with the wasm_shim compile target mapped as a volume
* Install authorino-operator
* Deploy authorino + certs based on `TLS_ENABLED`
* Deploy limitador
* Deploy envoyproxy with the local volume mapped
* Deploys talker-api
* Deploys simple authconfig and secret for the API key user-guide

To deploy all of the above:
```sh
make local-setup
```

Cleanup:
```sh
make local-cleanup
```

To test changes to the wasm binary you can simply run the following to redeploy in the cluster:

```sh
make build local-rollout
```

To watch the envoy + wasm logs:
```sh
```

To test the API:
```sh
```

```sh
curl -H "Host: test.a.auth.com" http://127.0.0.1:8000/get -i
```

```sh
curl -H "Host: test.a.rlp.com" http://127.0.0.1:8000/get -i
```

```sh
curl -H "Host: test.b.rlp.com" http://127.0.0.1:8000/get -i
```

```sh
curl -H "Host: test.c.rlp.com" -H "x-forwarded-for: 127.0.0.1" -H "My-Custom-Header-01: my-custom-header-value-01" -H "x-dyn-user-id: bob" http://127.0.0.1:8000/get -i
```

```sh
make local-cleanup
```


utils/deploy/envoy-notls.yaml
```diff
@@ -0,0 +1,269 @@
...
+kind: ConfigMap
+metadata:
+  labels:
+    app: envoy
+  name: envoy
+data:
+  envoy.yaml: |
+    static_resources:
+      clusters:
+        - name: authorino_wasm
```
adam-cattermole, line 11: When this cluster is named `authorino` I am unable to reach it from within the shim so for now it is named `authorino_wasm` - todo: investigate why this is the case

eguzki (changes requested):
Overall looks good.

* Documentation is needed
* I like the docker-compose environment, looks like more lightweight and quicker to work with. However, with authorino, due to it's complexity, and being a k8s controller, docker-compose is no longer an option. Should we keep both environments? any thoughts on this?

utils/kind/cluster.yaml
```diff
@@ -0,0 +1,9 @@
+---
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+  - role: control-plane
+    image: kindest/node:v1.30.0
+    extraMounts:
+      - hostPath: $(WASM_PATH)
```
eguzki, line 8: this file is a template, which is rendered with the right wasm path. This adds some (minor) extra complexity. 

Have you considered volume source being a secret which could be easier to manage and monitor?

adam-cattermole: My idea was to reduce the number of changes required to iterate as developing - so a simple rebuild of the `make build` has generated a new file in the kind cluster.

That being said it's still pretty simple to re-create the secret each time we rebuild so I'm not tied to one approach or the other. If you prefer the secret I can look into it

adam-cattermole:
@eguzki agreed, will look to add all the docs on using this.

> I like the docker-compose environment, looks like more lightweight and quicker to work with. However, with authorino, due to it's complexity, and being a k8s controller, docker-compose is no longer an option. Should we keep both environments? any thoughts on this?

I was also unsure on this one - I agree docker-compose is a good lightweight alternative for limitador only, but then again two different dev environments might have differences which make it hard to spot issues. I think that whether we remove the docker-compose or not I'll leave it to a future PR outside of this one though.

eguzki:
@adam-cattermole The PR includes some commits not related to the integration development environment. Not sure that is intended or not. 

Regarding the integration development environment, I think it needs  few more iterations, but we can polish it in following up PR's. LGTM.

adam-cattermole:
@eguzki Oh you're right, it's diverged from base branch, let me rebase again

eguzki (approved):

didierofrivia (approved):
🆒

PR #68, opened 2024-08-16
Add initial implementation of auth service

adam-cattermole:


adam-cattermole:
Woops wrong remote, will reopen

PR #69, opened 2024-08-16
Add initial implementation of auth service

adam-cattermole:
### Changes

- Refactor of the ratelimit service and auth services into a generic `GrpcServiceHandler`
- Addition of the new `HeaderResolver` which is passed into each `GrpcServiceHandler` to ensure the headers are only retrieved once (and set in a `OnceCell`)
- For now the Auth and RL services are only used to create the messages and could be moved to an alternative location (for now auth marked as dead code until used)
- Fixed the integration test order of hostcalls and removed for the tests where there are no longer retrieval of headers


src/service/auth.rs
```diff
@@ -0,0 +1,122 @@
...
+        http.set_headers(headers);
+        request.set_time(get_attribute("request.time").map_or(
+            Timestamp::new(),
+            |date_time: DateTime<FixedOffset>| Timestamp {
+                nanos: date_time.nanosecond() as i32,
+                seconds: date_time.second() as i64,
+                unknown_fields: Default::default(),
+                cached_size: Default::default(),
+            },
+        ));
```
adam-cattermole, line 87: I need to investigate how this field is expected to be set as this implementation is incorrect

adam-cattermole: I'm going to iterate on this at a later date

didierofrivia (approved):
🍸

PR #77, opened 2024-08-27
Operation dispatcher FSM

didierofrivia:

This PR introduces the _OperationDispatcher_ FSM, that it's responsible of not only keeping track of the current state of the operations, but also triggering the actual actions defined by configuration.
It also includes a refactor that was needed in order to wire the Filter, Services, ServiceHandlers and Operations together.

### The implementation:

* `OperationDispatcher`:
  *  its responsible of building the operations and exposing the `next` function in charge of triggering the current 
     `Operation` procedure
  * The `next` fn will return the current `Operation` when called.
* `Operation`:
  * Exposes its `State`, `Result` and `Extension` 
  *  will _trigger_ the request for the Auth/RateLimit GRPC service 
  * It will also track and mutate the `State` (`Pending` -> `Waiting` -> `Done`)
  * Responsible of providing the `hostcalls` fns
  
  
### Regarding the refactoring of the other pieces:
  
  * A new type `GrpcMessage` was created, implementing the trait `Message` in order to abstract its logic
  * `GrpcService` now keeps a `Rc` of `Extension`
  * `GrpcServiceHandler.send` method takes as args `get_map_values_bytes` and `grpc_call` following DI pattern in order to unit test the operations.
  * Configuration `Actions` have been added.
  
### Notes
* The wiring only replicates the previous behaviour, still needs to implement the actual flow with multiple actions
* The `Rules` from the `PluginConfiguration` follows still the old state, duplicating `DataType` that resides within the `Actions`
* Missing tests for the `service` module
* **FOR THE REVIEWERs OF THIS PR**: Don't go per commit, just go for the [Files Changed](https://github.com/Kuadrant/wasm-shim/pull/77/files) unless you want to go through a convoluted collection of changes.


src/service.rs
```diff
@@ -1,45 +1,174 @@
...
+        GrpcMessage::default()
+    }
+
+    fn default_instance() -> &'static Self
+    where
+        Self: Sized,
+    {
+        #[allow(non_upper_case_globals)]
+        static instance: ::protobuf::rt::LazyV2<GrpcMessage> = ::protobuf::rt::LazyV2::INIT;
+        instance.get(|| GrpcMessage::RateLimit(RateLimitRequest::new()))
```
alexsnaps, line 126: The requirement for this confuses me... couldn't this "just" be a factory function and be passed around as a fn pointer instead? Not a big deal tho

src/operation_dispatcher.rs
```diff
@@ -0,0 +1,291 @@
...
+            .borrow()
+            .first()
+            .map(|operation| operation.get_state().clone())
+    }
+
+    pub fn get_current_operation_result(&self) -> Result<u32, Status> {
+        self.operations.borrow().first().unwrap().get_result()
+    }
+
+    pub fn next(&self) -> Option<Operation> {
```
alexsnaps, line 146: I think this signature will need to be somewhat rethought... Probably something along the lines of a 

 - `.start()` that starts a "sequence", `initial` state
 - `.next(...)` with some params (gRPC response's `token_id` & `response_size`?) to go from one operation to the next
 - `()` i.e. "void"/unit, i.e. nothing left to do, representing the `terminated` state

alexsnaps (approved):
Left some notes, but I think this is good starting point... Let's iterate!

src/service.rs
didierofrivia: This one was to please the compiler, got the implementation from the message lib xD

src/operation_dispatcher.rs
didierofrivia: Yup, after discussing about how would it be consumed `on_grpc_response` and the requirements to digest the actual response needing the `response_size` makes sense. Will shoot a following PR addressing this. Thanks!

eguzki:
I want to understand and follow. Shouldn't the distpatcher be called on the `on_grpc_call_response` method of the Context trait?

didierofrivia:
@eguzki  you're correct, it should be call on `on_grpc_call_response` too, this PR is implementing the FSM but not changing the current behaviour. Next PR will address that since the API needs to change a bit, around this https://github.com/Kuadrant/wasm-shim/pull/77#discussion_r1745783219

eguzki:
ok, let's go then. Forget about the clippy thing for now.

eguzki (approved):
let's iterate! 🔨

PR #80, opened 2024-09-06
Clippy Allowing warnings

didierofrivia:


.github/workflows/rust.yaml
```diff
@@ -71,4 +71,4 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-targets --all-features -- -D warnings
```
eguzki, line 74: This is opening the door for all the warnings. Today have none due to this. 

Are we sure we want this?

didierofrivia: The alternative is to pin `rustc` to `1.80`, the one that is marking the allowing of `box_pointers` is `rustc 1.81.0 (eeb90cda1 2024-09-04)`

eguzki: What about  updating `src/envoy` path content? I am aware it is auto-generated from protobuf descriptors. So maybe change the source of truth or add some customization https://github.com/Kuadrant/wasm-shim/blob/main/build.rs#L55-L106

It is all about deleting 

```rust
#![allow(box_pointers)]
```

That is a lint being removed in rust 1.80

alexsnaps:
Whichever route you all decide to take, I'd like for it to remain _a temporary fix_ until [this PR](https://github.com/stepancheg/rust-protobuf/pull/733) is released into a newer version of the crate... thanks 🙏

didierofrivia:
@eguzki Added a `TODO` comment pointing out the PR that needs to be released

PR #81, opened 2024-09-09
On grpc response flow

didierofrivia:

This PR is yet another step closer to the new way the WASM Shim will be handling multiple operations of External Rate Limit and Auth.

It introduces the `GrpcMessageResponse` that will be responsible of parsing the response back from the external services into the correct type, thus allowing the Filter to decide on the flow given the different actions defined in configuration.

So far, it only replicates the previous Rate Limiting flow, opening the door to implement the Auth one.

Next step: Implement the actual Auth flow within the Filter.

Notes:

* Following this approach, it doesn't delegate the response parsing to the Operation Dispatcher, but to the Service GrpcMessageResponse, being the Filter shaping the flow of the actions and the Operations in charge of only dispatching them.


src/envoy/mod.rs
```diff
@@ -37,7 +37,7 @@ pub use {
         AttributeContext_Request,
     },
     base::Metadata,
-    external_auth::CheckRequest,
+    external_auth::{CheckRequest, DeniedHttpResponse, OkHttpResponse},
```
alexsnaps, line 40: I see this is more than RL then... which is fine ;)

src/filter/http_context.rs
```diff
@@ -124,55 +158,42 @@ impl Context for Filter {
...
+                ExtensionType::RateLimit => self.process_ratelimit_grpc_response(res, failure_mode),
             }
+
+            if let Some(_op) = self.operation_dispatcher.next() {
+            } else {
+                self.resume_http_request()
+            }
+        } else {
+            warn!("No Operation found with token_id: {token_id}");
+            self.handle_error_on_grpc_response(&FailureMode::Deny); // TODO(didierofrivia): Decide on what's the default failure mode
```
alexsnaps, line 196: Yeah, interesting... I think this _could_ (not asking to change the current behavior really) be a `.next().expect("State machine is broken, we expected an operation to handle the gRPC response")`, which if I'm not mistaken result in a `500`.

src/filter/http_context.rs
```diff
@@ -124,55 +158,42 @@ impl Context for Filter {
...
+                        "failed to parse grpc response body into GrpcMessageResponse message: {e}"
+                    );
+                    self.handle_error_on_grpc_response(failure_mode);
+                    return;
                 }
+            };
+            match operation.get_extension_type() {
+                ExtensionType::Auth => {} // TODO(didierofrivia): Process auth grpc response.
+                ExtensionType::RateLimit => self.process_ratelimit_grpc_response(res, failure_mode),
             }
```
alexsnaps, line 188: I think much of that logic could be encapsulated in the `operation` itself... Probably would need a bigger refactoring tho (e.g. an `enum` for how to transition, i.e. `()` or `GrpcMessageResponse`, to the next state)... Same with the code in `process_ratelimit_grpc_response`... its back & forth between these two layers that I don't think adds value, nor is required...

alexsnaps (approved):
Couldn't spot anything explicitly broken... So I'm in favor in merging. I'd then slightly focus on merging this all to `main` & worst case nightly catches something, and we have plenty of time to then fix it (while making our test coverage better :) 
That would require looking into the config bits on the kuadrant operator end tho

PR #87, opened 2024-09-20
Refactor pluginconfig

adam-cattermole:
## Changes

* Moved the `domain` field into the actions and renamed `scope`. This is the `host` for auth ContextExtensions and `domain` for ratelimit
* Started using the `data` field in the actions and removed the `data` field in rules

Example:

```yaml
---
extensions:
  limitador:
    type: ratelimit
    endpoint: limitador-cluster
    failureMode: deny
policies:
  - name: rlp-ns-A/rlp-name-A
    hostnames:
      - '*.toystore.com'
      - example.com
    rules:
      - conditions:
          - selector: request.path
            operator: eq
            value: /admin/toy
          - selector: request.method
            operator: eq
            value: POST
          - selector: request.host
            operator: eq
            value: cars.toystore.com
    actions:
      - extension: limitador
        scope: rlp-ns-A/rlp-name-A
        data:
          - static:
              key: rlp-ns-A/rlp-name-A
              value: "1"
          - selector:
              selector: auth.metadata.username
```

#### In progress

* [x] Move actions into the rules (merged from @didierofrivia in #88)
* [x] Store dynamic metadata from auth in the `filter_state` so it can be retrieved by other actions
* [x] Build the ratelimit descriptors at action time rather than beforehand so we can retrieve aforementioned `filter_state`

## Verification:

Use the new guide in the README to verify multi-action setup:

```
```

Port-forward envoy and watch the logs for all services:
```
kubectl logs -f deployment/authorino
kubectl logs -f deployment/limitador
```

Test the authenticated rate limiting:
```
curl -H "Host: test.a.multi.com" http://127.0.0.1:8000/get -i
# HTTP/1.1 401 Unauthorized
```

Alice has 5 requests per 10 seconds:
```
while :; do curl --write-out '%{http_code}\n' --silent --output /dev/null -H "Authorization: APIKEY IAMALICE" -H "Host: test.a.multi.com" http://127.0.0.1:8000/get | grep -E --color "\b(429)\b|$"; sleep 1; done
```

Bob has 2 requests per 10 seconds:
```
while :; do curl --write-out '%{http_code}\n' --silent --output /dev/null -H "Authorization: APIKEY IAMBOB" -H "Host: test.a.multi.com" http://127.0.0.1:8000/get | grep -E --color "\b(429)\b|$"; sleep 1; done
```

### DIFF
--- a/src/policy.rs
+++ b/src/policy.rs
@@ -1,24 +1,9 @@
-use crate::attribute::Attribute;
-use crate::configuration::{Action, DataItem, DataType, PatternExpression};
-use crate::envoy::{RateLimitDescriptor, RateLimitDescriptor_Entry};
-use crate::filter::http_context::Filter;
-use log::debug;
-use proxy_wasm::traits::Context;
+use crate::configuration::{Action, PatternExpression};
 use serde::Deserialize;
 
 #[derive(Deserialize, Debug, Clone)]
-#[serde(rename_all = "camelCase")]
-pub struct Condition {
-    pub all_of: Vec<PatternExpression>,
-}
-
-#[derive(Deserialize, Debug, Clone)]
 pub struct Rule {
-    //
-    #[serde(default)]
-    pub conditions: Vec<Condition>,
-    //
-    pub data: Vec<DataItem>,
+    pub conditions: Vec<PatternExpression>,
 }
 
 #[derive(Deserialize, Debug, Clone)]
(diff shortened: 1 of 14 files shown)


src/policy.rs
```diff
@@ -1,31 +1,15 @@
...
-}
-
 #[derive(Deserialize, Debug, Clone)]
 pub struct Rule {
-    //
-    #[serde(default)]
-    pub conditions: Vec<Condition>,
-    //
-    pub data: Vec<DataItem>,
+    pub conditions: Vec<PatternExpression>,
```
adam-cattermole, line 6: @eguzki preferred retaining the name `allOf` over `conditions` here to be explicit that all conditions must apply. I myself am of the opinion that `allOf` suggests there's an `anyOf` so prefer conditions, so would like other thoughts

adam-cattermole:
There's an issue with our design here, we've decoupled the rules from the data and actions, which makes some configurations impossible. We likely should nest the actions within the rules - if rule applies, actions performed with relevant data.

src/policy.rs
eguzki: I think it will eventually (very soon) be called `predicates`, right @alexsnaps ?

eguzki:
related https://github.com/Kuadrant/kuadrant-operator/issues/867

src/policy.rs
alexsnaps: I think whether these are `Predicate`s or `PatternExpression`s here is orthogonal to @adam-cattermole 's point. I _think_ he wants to keep the field names `allOf`, which indeed eventually will be something else than `PatternExpression`. It would eventually become a cel expression representing a [`Predicate`](https://en.wikipedia.org/wiki/Propositional_function) indeed, tho, as discussed, how that would look like is probably dependent on what work the Kuadrant Operator does with the initial unparsed expression... if it is parsed entirely, we could inject the AST straight into here (more so if there is an easy way to (de)serialize it to yaml, from Go to Rust) or just pass the `String` around...

adam-cattermole:
I think we need some follow ups to improve performance here, will look to iterate to make things better.. we're also in need of new integration tests for auth, as well as auth + rl.. but it's working

didierofrivia (approved):
Great, from here on we can dedicate ourself to do the missing integration tests for auth, multi extension and make it faster/prettier. 💪🏼

PR #88, opened 2024-09-25
Refactoring actions

didierofrivia:
This PR is yet another step closer to get https://github.com/Kuadrant/wasm-shim/issues/58 done.

Now `Actions` reside within a given `Rule`. The logic behind this change is that actions are dependant of the rule that matches, and not a set of actions for all of them.

* Only 1 Rule applies and provides a set of actions
* The building of the descriptors needed for RL is not relying on `Filter` anymore
* Also, the rule matching and descriptors use `hostcalls::get_property` directly

Notes:
* Since we are not passing down the `Filter` anymore for building the descriptors, we are not logging the `context_id` when getting de properties.

### DIFF
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -462,10 +462,12 @@ impl TryFrom<PluginConfiguration> for FilterConfig {
 
         for rlp in config.policies.iter() {
             for rule in &rlp.rules {
-                for datum in &rule.data {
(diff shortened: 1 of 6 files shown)


didierofrivia:
Closing this one, changes already migrated to https://github.com/Kuadrant/wasm-shim/pull/87

PR #90, opened 2024-10-01
Refactoring operation dispatcher

didierofrivia:
This PR cleans a bit the work done for the operation dispatcher. Among the changes:

* `Operations` store its status and result in RefCell for interior mut
* `OperationDispatcher` keeps a `Vec` of `Rc<Operation>`, then indexes cloning the `Rc` **instead** of cloning the entire object.
* Removed `get_current_operation_result` fn that wasn't used.


adam-cattermole (approved):
Nice, lgtm

PR #100, opened 2024-10-04
remove docker compose environment

eguzki:
### what

* Remove docker compose environment
* enable kind based environment to run with either `release` or `debug` build mode (by default is `debug`)
```
make local-setup BUILD=[release|debug] 
```


adam-cattermole (approved):
Nice, all looks great!

Issue #93, opened 2024-10-02
Make the timeout of the rate limit request configurable


roivaz:
The current timeout value is `5s` and not configurable (see [code](https://github.com/Kuadrant/wasm-shim/blob/main/src/service.rs#L102)). Also, 5 seconds seems way too high. A more sensible default could be `20ms`, same as the envoy [ratelimit filter](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ratelimit/v3/rate_limit.proto#envoy-v3-api-msg-extensions-filters-http-ratelimit-v3-ratelimit) uses.

alexsnaps:
To be considered whether we'd want 2 of these, or say one "per service", as I could see Authorino needing a different value than Limitador

adam-cattermole:
I'd say that we include it as part of the [extension configuration](https://github.com/Kuadrant/wasm-shim/blob/main/src/configuration.rs#L531-L537) in the wasm-shim so that it's configurable by service, the same way as the failure mode.

PR #101, opened 2024-10-04
Timeout configurable

didierofrivia:
Closes https://github.com/Kuadrant/wasm-shim/issues/93

**Example config:**

```yaml
extensions:
  limitador:
    type: "ratelimit",
    endpoint: "limitador-cluster",
    failureMode: "deny",
    timeout: "5s"
policies:
...
```

The timeout value must be string, and must be in the format of a duration. Default to `20ms`

**Examples of valid timeout values:**
* 1ms parses as 1 millisecond
* 1.5ms parses as 1 millisecond and 500 microseconds
* 5s parses as 5 seconds

### DIFF
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -675,13 +675,13 @@ mod test {
             "authorino": {
                 "type": "auth",
                 "endpoint": "authorino-cluster",
-                "failureMode": "deny"
+                "failureMode": "deny",
                 "timeout": "24ms"
             },
             "limitador": {
                 "type": "ratelimit",
                 "endpoint": "limitador-cluster",
-                "failureMode": "allow"
+                "failureMode": "allow",
                 "timeout": "42ms"
             }
         },
@@ -753,7 +753,7 @@ mod test {
             assert_eq!(auth_extension.extension_type, ExtensionType::Auth);
             assert_eq!(auth_extension.endpoint, "authorino-cluster");
             assert_eq!(auth_extension.failure_mode, FailureMode::Deny);
-            assert_eq!(auth_extension.timeout, Timeout(Duration::from_millis(24)));
+            assert_eq!(auth_extension.timeout, Timeout(Duration::from_millis(24)))
         } else {
             panic!()
         }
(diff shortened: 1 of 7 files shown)


adam-cattermole (approved):
Nice job, this seems to be working well! I also verified by changing one of the integration test configs and could see the timeout in the grpc request change. Just needs a rebase I think

src/configuration.rs
```diff
@@ -534,6 +536,52 @@ pub struct Extension {
...
+    {
+        self.visit_string(String::from(string))
+    }
+
+    fn visit_string<E>(self, string: String) -> Result<Self::Value, E>
+    where
+        E: Error,
+    {
+        match duration(Arc::new(string)) {
+            Ok(Value::Duration(duration)) => Ok(Timeout(duration.to_std().unwrap())),
```
alexsnaps, line 580: I think this could be an issue here, if I understand this right a "`CelDuration`" can be negative, a std can't if I'm not mistaken. So we expose it as if it could, then would panic if one would provide a negative one, which would be nonsensible in any case tho...

alexsnaps (approved):
lgtm, modulo that one comment on the negative durations. But this is internal again, so not a big deal, tho I'd create an issue if we don't fix it now in place. Thanks 🙏

src/configuration.rs
```diff
@@ -534,6 +536,52 @@ pub struct Extension {
...
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        deserializer.deserialize_str(TimeoutVisitor)
+    }
+}
+
+struct TimeoutVisitor;
+impl<'de> Visitor<'de> for TimeoutVisitor {
```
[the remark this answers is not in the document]
alexsnaps: The default impl. in the trait [eg. here](https://rust-lang.github.io/hashbrown/src/serde/de/mod.rs.html#1484), got it!

src/configuration.rs
didierofrivia: you are correct, this would panic on a negative as `OutOfRangeError(())`

PR #102, opened 2024-10-07
Fix calling next operation when error

didierofrivia:


src/service/rate_limit.rs
```diff
@@ -70,8 +72,9 @@ impl RateLimitService {
                     )
                     .unwrap()
                 });
+                Ok(())
             }
-            _ => {}
+            _ => Err(StatusCode::InternalServerError),
```
adam-cattermole, line 77: I'm wondering about this case here, this is the fall through for the matches above - if we always expect the above should we do `GrpcService::handle_error_on_grpc_response(failure_mode);` as well, similar to the auth case?

adam-cattermole (approved):
Minor suggestion but otherwise changes look good

PR #107, opened 2024-10-08
Process headers_to_remove

adam-cattermole:


src/service/auth.rs
```diff
@@ -102,6 +102,17 @@ impl AuthService {
...
                 Some(CheckResponse_oneof_http_response::ok_response(ok_response)) => {
                     debug!("process_auth_grpc_response: received OkHttpResponse");
+                    ok_response
+                        .get_headers_to_remove()
+                        .iter()
+                        .for_each(|header| {
+                            hostcalls::set_map_value(
+                                MapType::HttpResponseHeaders,
+                                header.as_str(),
+                                None,
```
adam-cattermole, line 112: If value is `None` it calls `proxy_remove_header_map_value`

Issue #108, opened 2024-10-11
Config revamp and purpose


alexsnaps:
We have a few issues with the way we configure our code to do the proper thing based of how a `Policy` gets attached to Gateway API objects. Here is a tentative at a proposal to refine the config to a bare minimal of what wasm is supposed to do within the Gateway.

## What is wasm even need for?

Upon serving a request, our wasm module needs to decide whether some actions need to be performed. This happens in "conceptual" steps: 
 - Are there any `Policy` attached (directly or indirectly) to the `HTTPRouteMatch`? 
 - Are the conditions of that `Policy` met?
 - If an action is taken, how does it affect the request?

The first step is to evaluate whether the request matches any of `HTTPRouteMatch` _which do have a `Policy` attached_, otherwise there is no action to be taken.

### Mapping `HTTPRouteRule` to actual requests

In order to map a request to an `HTTPRouteMatch`, we need to evaluate which is matched by the request. There is [a whole section](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteRule) in the Gateway API specification that describes the precedence rules for matching requests. Ideally the configuration would have the list of `Actions` to be sorted accordingly. Simplistically, we'd just "fall through" the list to find the best (i.e. the first) match. Obviously that list would contain only  `HTTPRouteMatch` for which actions are to be taken (i.e. ones for which at least one policy has been attached to). That "`ActionSet`" to be taken would be identified by these "`RouteRuleConditions`", composed of whatever the `HTTPRouteMatch` expresses, added to the `Hostname`s defined by both the `HTTPRouteSpec` and the `GatewaySpec`'s `Listener`s.

That first step would be the `RouteRuleConditions` expressed as "top-level" conditions in the order list of `ActionSet`.

### Evaluating whether an action is to be taken

`Action`s within that `ActionSet` are themselves conditional. These `Condition`s are expressed by the user as part of their `Policy` definition (e.g. only apply rate limiting if the request isn't from a known set of IP addresses, like our offices).

Each individual `Action` has first its `Condition`s evaluated and only upon them being satisfied, it is invoked. Otherwise, the next `Action` of the `ActionSet` gets evaluated for invocation (e.g. _Authorino_). Further down the `ActionSet`s, `Condition`s can reuse data from the previous stage, e.g. only limit "anonymous" traffic. 

[...]

Issue #104, opened 2024-10-07
Extend wasm configuration for conditional multiple descriptors for rate limiting


eguzki:
### Limitations

* Currently, for rate limiting, the action `data` can only generate one descriptor. 
```yaml
actions:
      - extension: ratelimit-ext
        scope: rlp-ns-A/rlp-name-A
        data:
          - selector:
              selector: request.headers.My-Custom-Header
          - static:
              key: admin
              value: "1"
```
Additionally, only top level conditions used to conditionally include descriptor entries. And it is all or nothing. 

* Only one rule can be activated per request. This is not a limitation *per se*. It is a limitation taking into account how kuadrant's control plane generates wasm configuration. For example, the following RLP targeting a simple HTTPRoute (routing traffic for `GET /get`:

```yaml
  limits:
    limitA:
      rates:
        - duration: 10
          limit: 8
          unit: second
    limitB:
      rates:
        - duration: 5
          limit: 3
          unit: second
```

generates the following wasm configuration:

```yaml
    rules:
    - conditions: # RULE for limit A
      - allOf:
          - operator: startswith
             selector: request.url_path
             value: /get
      actions:   
      - extension: ratelimit-ext
        scope: rlp-ns/rlp-name
        data:
        - static:
             key: limit.limitA__6fb66d73
             value: '1'
    - conditions: # RULE for limit B
      - allOf:
          - operator: startswith
             selector: request.url_path
             value: /get
      actions:   
      - extension: ratelimit-ext
        scope: rlp-ns/rlp-name
        data:
        - static:
             key: limit.limitB__6fb66d73
             value: '1'
```

This configuration will always activate the first rule making the second one shadowed (disabled effectively). However, the expected behavior of the user is to have both counters being increased. That requires that the descritor(s) sent would include ` key : "limit.limitA__6fb66d73", value: "1"` **and**  ` key : "limit.limitB__2726263", value: "1"`. However, the current implementation does not send both descriptor entries, only the one of the first rule. In practice, this leads to some limit to be enforced and not the other. 

This is not a limitation *per se*, because the wasm config could be like the following and the desired behavior (activating multiple counters) would be achieved.

```yaml
    rules:
          - actions:
              - data:
                  - static:
                      key: limit.limitA__6fb66d73
                      value: '1'
                  - static:
                      key: limit.limitB__2726263
                      value: '1'
                extension: limitador
                scope: kuadrant/rlp-name-a
            conditions:
              - allOf:
[...]

PR #109, opened 2024-10-11
Store multiple policies in the index

adam-cattermole:
Prior to making more substantial changes to the configuration I've quickly made the change to store multiple policies for each hostname:

* We now iterate over each "policy" at the hostname of the index, finding the first one where the "rules" apply.
* Updated the log ordering in tests - we now debug log when checking the conditions prior to knowing we've selected a policy, so the "policy selected" statement comes after

### DIFF
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -466,8 +466,8 @@ impl TryFrom<PluginConfiguration> for FilterConfig {
     fn try_from(config: PluginConfiguration) -> Result<Self, Self::Error> {
         let mut index = PolicyIndex::new();
 
-        for rlp in config.policies.iter() {
-            for rule in &rlp.rules {
+        for policy in config.policies.iter() {
+            for rule in &policy.rules {
                 for condition in &rule.conditions {
                     for pe in &condition.all_of {
                         let result = pe.compile();
@@ -486,8 +486,8 @@ impl TryFrom<PluginConfiguration> for FilterConfig {
                 }
             }
 
-            for hostname in rlp.hostnames.iter() {
-                index.insert(hostname, rlp.clone());
+            for hostname in policy.hostnames.iter() {
+                index.insert(hostname, Rc::new(policy.clone()));
             }
         }
 
@@ -1163,15 +1163,17 @@ mod test {
 
         let result = FilterConfig::try_from(res.unwrap());
         let filter_config = result.expect("That didn't work");
-        let rlp_option = filter_config.index.get_longest_match_policy("example.com");
+        let rlp_option = filter_config
+            .index
+            .get_longest_match_policies("example.com");
         assert!(rlp_option.is_some());
 
         let rlp_option = filter_config
             .index
-            .get_longest_match_policy("test.toystore.com");
+            .get_longest_match_policies("test.toystore.com");
         assert!(rlp_option.is_some());
 
-        let rlp_option = filter_config.index.get_longest_match_policy("unknown");
+        let rlp_option = filter_config.index.get_longest_match_policies("unknown");
         assert!(rlp_option.is_none());
     }
 
--- a/src/filter/http_context.rs
+++ b/src/filter/http_context.rs
@@ -29,8 +29,13 @@ impl Filter {
         }
     }
 
-    fn process_policy(&self, policy: &Policy) -> Action {
-        if let Some(rule) = policy.find_rule_that_applies() {
+    fn process_policies(&self, policies: &[Rc<Policy>]) -> Action {
+        if let Some(rule) = policies.iter().find_map(|policy| {
+            policy.find_rule_that_applies().map(|rule| {
+                debug!("#{} policy selected {}", self.context_id, policy.name);
+                rule
+            })
+        }) {
             self.operation_dispatcher
                 .borrow_mut()
                 .build_operations(rule);
@@ -66,7 +71,7 @@ impl HttpContext for Filter {
         match self
             .config
             .index
-            .get_longest_match_policy(self.request_authority().as_str())
+            .get_longest_match_policies(self.request_authority().as_str())
         {
             None => {
                 debug!(
@@ -75,10 +80,7 @@ impl HttpContext for Filter {
                 );
                 Action::Continue
             }
-            Some(policy) => {
-                debug!("#{} policy selected {}", self.context_id, policy.name);
-                self.process_policy(policy)
-            }
+            Some(policies) => self.process_policies(policies),
         }
     }
 
--- a/src/filter/http_context.rs
+++ b/src/filter/http_context.rs
@@ -29,6 +29,7 @@ impl Filter {
         }
     }
 
+    #[allow(clippy::manual_inspect)]
     fn process_policies(&self, policies: &[Rc<Policy>]) -> Action {
         if let Some(rule) = policies.iter().find_map(|policy| {
             policy.find_rule_that_applies().map(|rule| {
--- a/src/policy.rs
+++ b/src/policy.rs
@@ -16,7 +16,7 @@ pub struct Rule {
     pub actions: Vec<Action>,
 }
 
-#[derive(Deserialize, Debug, Clone)]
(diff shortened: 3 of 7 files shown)


eguzki:
For the record, I want to write here that I would not follow the path opened by this PR.  Let me try my (last) bullet and if still any of you do not buy this, I will accept it and approve this.

TL;DR: let's not reinvent the wheel here. 

Envoy, for routing, also is doing a similar task as the wasm needs to do. Envoy has a set of routes indexed by `domains` and some matchers to select one route out of multiple routes for a given domain. This is what Envoy calls [VirtualHost](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-virtualhost). And each virtualhost object, has a list of `domains`.

```yaml
virtual_hosts:
  - name: gateway-system/kuadrant-ingressgateway/http/api_toystore_com
    domains:
      - api.toystore.com
```

It turns out that `domains` field, as explained in their [doc](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-virtualhost), 

```
The longest wildcards match first. Only a single virtual host in the entire route configuration can match on *. A domain must be unique across all virtual hosts or the config will fail to load.
```

And, more importantly, the domain search order is as follows (also from the doc). 

```
        Exact domain names: www.foo.com.

        Suffix domain wildcards: *.foo.com or *-bar.foo.com.

        Prefix domain wildcards: foo.* or foo-*.

        Special wildcard * matching any domain.
```

In conclusion: the [order of virtual_hosts does **not** matter](https://github.com/envoyproxy/envoy/issues/18874).

What are we doing in this PR? We are making the order of policies (the virtualhosts of our Wasm configuration) to matter. Moreover, we are making the configuration far more complicated to understand as we are allowing multiple "policies" under the same domain. It is totally counter intuitive IMO. And Wasm policy rules are meant to do that precisely.

I am advocating for following Envoy's approach. In the first place, I would implement, in Wasm configuration terms:

```
A hostname must be unique across all policies or the config will fail to load.
```

Currently, the Wasm-shim only implements `The longest wildcards match first`. I was very careful to have that functionality from the very beginning. But we should also change the trie implementation with something smarter about

```
        Exact domain names: www.foo.com.

        Suffix domain wildcards: *.foo.com or *-bar.foo.com.

        Prefix domain wildcards: foo.* or foo-*.

        Special wildcard * matching any domain.
```

So, instead of allowing the wasm configuration enabled by this PR, which is something like this:

```yaml
      policies:
      - hostnames:
        - '*'
        name: gateway-system/gw
        rules:
        - actions:
          - data:
            - static:
                key: limit.limit1__e3f1f251
                value: "1"
            extension: limitador
            scope: gateway-system/A
          conditions:
          - allOf:
            - operator: eq
              selector: request.url_path
              value: /toys
            - operator: eq
              selector: request.method
              value: GET
      - hostnames:
        - '*'
        name: gateway-system/gw
        rules:
        - actions:
          - data:
            - static:
                key: limit.limit1__e3f1f251
                value: "1"
            extension: limitador
            scope: gateway-system/B
          conditions:
          - allOf:
            - operator: startswith
              selector: request.url_path
              value: /
            - operator: eq
              selector: request.method
              value: GET
```

I suggest having something like:

```yaml
      policies:
      - hostnames:
        - '*'
        name: policyA
        rules:
        - name: ruleA
          actions:
          - data:
            - static:
                key: limit.limit1__e3f1f251
                value: "1"
            extension: limitador
            scope: gateway-system/A
          conditions:
          - allOf:
            - operator: eq
              selector: request.url_path
              value: /toys
            - operator: eq
              selector: request.method
              value: GET
        - name: ruleB
          actions:
          - data:
            - static:
                key: limit.limit1__e3f1f251
                value: "1"
            extension: limitador
            scope: gateway-system/B
          conditions:
          - allOf:
            - operator: startswith
              selector: request.url_path
              value: /
            - operator: eq
              selector: request.method
              value: GET
```

If the issue is tracking which policy and rule was activated, we can easily handle that adding `name` field to the rules as in the example above. Let's not create one "so-called-policy" per listener-HTTPRouteRule pair. Kuadrant's control plane can write HTTPRules for a given listener in the corresponding policy/rule pair. And still follow the same order of preference defined by the Gateway API. 

cc @alexsnaps

guicassolato:
_My 2 cents:_

I agree that the order of the configs should not matter. The rest I honestly (and deeply respectfully) believe is conflating API concern with implementation detail.

Indexing "policies" by hostname does not mean the config has to require unique hostnames. I can see how that could be important for one who writes an Envoy config, where possibly the mental model of the topology needs to fit the uniqueness constraint of the domain names. Arguably, that could be because Envoy, in this case, has decided to leak an implementation decision (indexation) to its API, thus shaping the way its users have to write configs.

The wasm shim is not Envoy though. In the whole scope of Kuadrant, users are not expected to write wasm configs themselves. Rather, they are invited to think in terms of Gateway API resources. How those resources are organised will dictate the mental model users will establish, to reason about the state of things, how requests will be handled.

Even without writing the wasm configs themselves and perhaps more even so because of that, having an API that reflects that mental model is IMO a good thing. In fact, I even wonder (again in a very respectful manner) why gateways in general wouldn't give this idea a shot.

alexsnaps:
>I agree that the order of the configs should not matter. The rest I honestly (and deeply respectfully) believe is conflating API concern with implementation detail.

I think we had agreed that the order would _matter_ and be ordered by the Kuadrant Operator. But these are _NOT_ virtual hosts, the things ordered are the `HTTPRouteRule`s, that happen to _also have a hostname associated. At this stage I'll be honest I'm getting slightly frustrated. We are running in circles _and not getting to any working software_. 

The order that matters _in this PR_ isn't the hostname's, but the "Rule", that's what is ordered. This PR _fixes the issue where when having 2 "so called policies" have the same hostname, instead of replacing the last one seen (in the config's order, _as we had decided_), it [appends it](https://github.com/Kuadrant/wasm-shim/pull/109/files#diff-b815c495fb396540c708add60a978c71cb6d3c5dbca473e284478f875904a79eR17-R25). Then, when looking up policies based on the hostname, [searches for the first policy that matches given the other attributes of the request](https://github.com/Kuadrant/wasm-shim/pull/109/files#diff-0f0acfa3a818202889e890aef9c5d971b3977ab3f2503c9bd9798c10effac811R32-R39). Again, as we had discussed.  

So someone tells me where this diverges from what was agreed upon. If now there are better ways to achieve the same, which I'm sure there is, as I consider this config utter convoluted for reasons that I fail to see... sure! But in the meantime this is _working software_!

eguzki (approved):
That being said, LGTM

guicassolato:
>> I agree that the order of the configs should not matter.
> I think we had agreed that the order would _matter_ and be ordered by the Kuadrant Operator.

Sorry, @alexsnaps. I wasn't clear in my comment before. I do not want to change anything here.

My comment was more in the general sense of, as a principle, maybe in a perfect world, users should not have to care about the order of their configs – and neither they will, while something else does that for them, whether it's the wasm module or the Kuadrant Operator. And yes, we agreed to be the latter, for reasons already discussed (e.g. identical source of truths whose precedence can only be known at the level of the operator that handles). In fact, I've already started working on the part of [sorting policies _first_ by hostname](https://github.com/Kuadrant/kuadrant-operator/pull/893/commits/6e04e348eb4d15709a9250009e98b0ba5e0aee78).

But as you rightfully pointed out, it's the ordering by _full matching rules_ (not only hostnames) that matters. I only focused on hostnames before because of the current structure of the config. For a better discussion about other ways to struct the config, we have #108.

> conflating API concern with implementation detail

That was in reference to the criticism raised against allowing multiple policies per hostname, not about the PR, which I'm totally support.

adam-cattermole:
I _think_ we're in agreement that this is moving the direction we had decided (albeit a patch on the old convoluted config vs a new one), so I'm going to merge and we can continue to discuss potential solutions in #104 and #108

PR #110, opened 2024-10-15
Config revamp

adam-cattermole:
### Changes

This is moving towards a configuration based on the suggestion in #108.

Currently this is looking as follows:

```yaml
services:
  auth-service:
    type: auth
    endpoint: auth-cluster
    failureMode: deny
    timeout: 10ms
  ratelimit-service:
    type: ratelimit
    endpoint: ratelimit-cluster
    failureMode: deny
actionSets:
  - name: rlp-ns-A/rlp-name-A
    routeRuleConditions:
      hostnames: [ "*.toystore.com" ]
      matches:
      - selector: request.url_path
        operator: startswith
        value: /get
      - selector: request.host
        operator: eq
        value: test.toystore.com
      - selector: request.method
        operator: eq
        value: GET
    actions:
    - service: ratelimit-service
      scope: rlp-ns-A/rlp-name-A
      conditions: []
      data:
      - selector:
          selector: request.headers.My-Custom-Header
      - static:
          key: admin
          value: "1"
```

I've separated all of the changes by commit to try make it easier to read the changes.

The primary change to non-config types is that we now have `actions.conditions` so we can conditionally perform an action - example provided in the envoy configuration that only calls the rate limiting service when `user_id == 'alice'`:

```
```

Port-forward envoy and watch the logs for all services:
```
kubectl logs -f deployment/authorino
kubectl logs -f deployment/limitador
```

Test unauthenticated does not get rate limited:
```
curl -H "Host: test.b.multi.com" http://127.0.0.1:8000/get -i
# HTTP/1.1 401 Unauthorized
```

Bob is not rate limited (note no calls to limitador in logs):
```
while :; do curl --write-out '%{http_code}\n' --silent --output /dev/null -H "Authorization: APIKEY IAMBOB" -H "Host: test.b.multi.com" http://127.0.0.1:8000/get | grep -E --color "\b(429)\b|$"; sleep 1; done
```

Alice has 5 requests per 10 seconds:
```
while :; do curl --write-out '%{http_code}\n' --silent --output /dev/null -H "Authorization: APIKEY IAMALICE" -H "Host: test.b.multi.com" http://127.0.0.1:8000/get | grep -E --color "\b(429)\b|$"; sleep 1; done
```

### Outstanding

- [x] Tests for conditional actions
- [x] Retrieve the `scope` from `action.data` (won't do)

For a future PR:
- Convert `routeRuleConditions.matches` and `action.conditions` from `PatternExpression` to CEL
- "Merge" subsequent actions of the same type to only make a single call for ratelimit

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ actionSets:
     actions:
     - service: ratelimit-service
       scope: rlp-ns-A/rlp-name-A
+      conditions: []
       data:
       - selector:
           selector: request.headers.My-Custom-Header
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ extensions:
     type: ratelimit
     endpoint: ratelimit-cluster
     failureMode: deny
-policies:
+actionSets:
   - name: rlp-ns-A/rlp-name-A
     hostnames: [ "*.toystore.com" ]
     rules:
@@ -163,7 +163,7 @@ To expose the envoy endpoint run the following:
 kubectl port-forward --namespace default deployment/envoy 8000:8000
 ```
 
-There is then a single auth policy defined for e2e testing:
+There is then a single auth action set defined for e2e testing:
 
 * `auth-a` which defines auth is required for requests to `/get` for the `AuthConfig` with `effective-route-1`
 
@@ -177,7 +177,7 @@ curl -H "Host: test.a.auth.com" -H "Authorization: APIKEY IAMALICE" http://127.0
 # HTTP/1.1 200 OK
 ```
 
-And some rate limit policies defined for e2e testing:
+And some rate limit action sets defined for e2e testing:
 
 * `rlp-a`: Only one data item. Data selector should not generate return any value. Thus, descriptor should be empty and
   rate limiting service should **not** be called.
--- a/README.md
+++ b/README.md
@@ -34,15 +34,15 @@ actionSets:
       - selector: request.method
         operator: eq
         value: GET
-      actions:
-      - extension: ratelimit-ext
-        scope: rlp-ns-A/rlp-name-A
-        data:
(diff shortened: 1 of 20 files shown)


eguzki (commented):
Looking good

eguzki:
Verification steps working

eguzki:
> Latest commit removes `action.scope` and this information is retrieved from the data section now

From `data`? This is unexpected. From a static data item with a key name (by convention) `domain`?

What if there are many, which one has preference?

adam-cattermole:
I've dropped that commit fwiw. I tend to agree that right now the purpose of the data and `scope` is different and the way (we currently) process the data items means we need some custom logic to process a specific keyed data item, which doesn't seem right

guicassolato:
> > Latest commit removes `action.scope` and this information is retrieved from the data section now
> 
> From `data`? This is unexpected. From a static data item with a key name (by convention) `domain`?

I agree with @eguzki. I don't like this change.

Whether as `domain` or as `authCfg`, in both cases this is a _required_ piece of data and in both cases it has the meaning of “scope”. Inside of `data`, it feels out of place IMO. Unlike the rest of what goes there, scope is not as much "user-defined" data.

The way it was before, we could technically have an action without any data, which makes sense to me since we have default data that comes from the type of action (`service`) anyway. In fact, `auth` is an example of action that typically would involve zero user-defined data. And I could speculate with meaning to that for `ratelimit` as well (e.g. letting Limitador knows about request without activating any limit).

Anyway, I honestly liked more how it was before. Thanks for ditching the commit!

eguzki (approved):

didierofrivia (approved):
💪🏼

PR #111, opened 2024-10-15
source.remote_address well known attribute

### LINKED ISSUE #97: remote_address
`remote_address` is the action name exposed by Envoy filter [envoy.filters.http.ratelimit](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/rate_limit_filter). The description of [remote_address](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-ratelimit-action-remoteaddress)

```
The following descriptor entry is appended to the descriptor and is populated using the trusted address from [x-forwarded-for](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#config-http-conn-man-headers-x-forwarded-for):

("remote_address", "<trusted address from x-forwarded-for>")
```

This seems pretty fundamental feature to rate limit by IP. 

The ask here is to create a *WellKnownAttribute* :tm: that reproduces the same behavior. 

### Related info
* Envoy `trusted client address` computation based on `X-Forwarded-For` header https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-for
* Nice article explaining “Real” IP address computation by Envoy https://www.zhaohuabing.com/post/2024-05-17-client-ip-en/
* Set `numTrustedProxies` using Istio https://istio.io/latest/docs/ops/configuration/traffic-management/network-topologies/#configuring-x-forwarded-for-headers
* Configure Client IP Detection using EnvoyGateway https://gateway.envoyproxy.io/latest/tasks/traffic/client-traffic-policy/#configure-client-ip-detection

eguzki:
## What

This PR introduces a new *Well Known Attribute* called `source.remote_address`. 

#### Why
Envoy's `source.address` (well known) [attribute](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes#connection-attributes) includes the port as `IP:PORT`. It is desirable to run rate limiting or auth policies relying only on the client address. Hence, this `source.remote_address` can be used to implement those policies. 

#### Attribute Value
This attribute evaluates to the `trusted client address` (IP address without port) as it is being defined by [Envoy Doc](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-for). 

In summary, the `trusted client address` is calculated as follows: By default, Envoy gateway instances are configured to append the source IP address of the immediate downstream node’s connection of the incoming HTTP connections to the `X-Forwarded-For` header. By default Envoy sets the number of trusted hops to 0, 
indicating that Envoy should use the address the connection is opened from, 
rather than a value inside the X-Forwarded-For list. 
Increasing this count will have Envoy use the `n`th value from the list, counting from the right.

#### Where can I use it?
This attribute can be used on `selector` fields.  

As conditions:

```yaml
- conditions:
  - allOf:
    - selector: source.remote_address
      operator: eq
      value: 50.0.0.1
``` 

As `SelectorItem` to generate descriptor entries. 

```yaml
data:
- selector:
    selector: source.remote_address
``` 

Fixes #97 

## Verification Steps

For the verification steps, this PR adds a new configuration to the existing local development environment. It's the new set of rules for `*.d.rlp.com`.

```yaml
{                                                
  "name": "rlp-ns-D/rlp-name-D",                 
  "hostnames": [                                 
    "*.d.rlp.com"                                
  ],                                             
  "rules": [                                     
     {                                            
          "conditions": [
               {
                    "allOf": [
                           {
                                "selector": "source.remote_address",
                                 "operator": "neq",
                                 "value": "50.0.0.1"
                             }
                     ]
                }
        ], 
       "actions": [                               
           {                                        
               "extension": "limitador",              
               "scope": "rlp-ns-D/rlp-name-D",        
               "data": [                              
                    {                                 
                         "selector": {                      
                                 "selector": "source.remote_address"
                          }                                  
                     }                                    
                 ]                                      
            }                                        
         ]                                          
      }                                            
   ]                                              
},                                               
```

And a new limit configuration

```yaml
- namespace: rlp-ns-D/rlp-name-D  
  max_value: 2                    
  seconds: 10                     
  conditions: []                  
  variables:                      
  - source.remote_address       
```

That configuration enables source based rate limiting on `*.d.rlp.com` subdomains, with only one "privileged" exception: IP "50.0.0.1" will not be rate limited. 

* Run the environment

```
```

Port-forward envoy:
```
```

Alice (IP: `40.0.0.1`) has 2 requests per 10 seconds:
```
while :; do curl --write-out '%{http_code}\n' --silent --output /dev/null  -H "X-Forwarded-For: 40.0.0.1" -H "Host: test.d.rlp.com" http://127.0.0.1:8000/get | grep -E --color "\b(429)\b|$"; sleep 1; done
```
Curl tool logs should show only two requests allowed for every 10 requests. 

Bob (IP: `30.0.0.1`) has 2 requests per 10 seconds:
```
while :; do curl --write-out '%{http_code}\n' --silent --output /dev/null  -H "X-Forwarded-For: 30.0.0.1" -H "Host: test.d.rlp.com" http://127.0.0.1:8000/get | grep -E --color "\b(429)\b|$"; sleep 1; done
```
Curl tool logs should show only two requests allowed for every 10 requests. 

Charley (the *Boss*) with privileged IP `50.0.0.1` does not get rate limited.

```
while :; do curl --write-out '%{http_code}\n' --silent --output /dev/null  -H "X-Forwarded-For: 50.0.0.1" -H "Host: test.d.rlp.com" http://127.0.0.1:8000/get | grep -E --color "\b(429)\b|$"; sleep 1; done
```
Curl tool logs should show all requests being accepted. 

> Note: running all the three clients also shows that rate limiting counter is not shared and they have their own counter.  IOW, each one has 2 requests every 10 seconds.

### DIFF
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -1122,7 +1122,6 @@ mod test {
                 selector: "request.id".to_string(),
                 operator: WhenConditionOperator::Equal,
                 value: "request_id".to_string(),
-                path: Default::default(),
                 compiled: Default::default(),
             };
             p.compile().expect("Should compile fine!");
@@ -1140,7 +1139,6 @@ mod test {
                 selector: "request.id".to_string(),
                 operator: WhenConditionOperator::Equal,
                 value: "\"request_id\"".to_string(),
-                path: Default::default(),
                 compiled: Default::default(),
             };
             p.compile().expect("Should compile fine!");
@@ -1158,7 +1156,6 @@ mod test {
                 selector: "request.id".to_string(),
                 operator: WhenConditionOperator::Equal,
                 value: "123".to_string(),
-                path: Default::default(),
                 compiled: Default::default(),
             };
             p.compile().expect("Should compile fine!");
@@ -1176,7 +1173,6 @@ mod test {
                 selector: "foobar".to_string(),
                 operator: WhenConditionOperator::Equal,
                 value: "\"123\"".to_string(),
-                path: Default::default(),
                 compiled: Default::default(),
             };
             p.compile().expect("Should compile fine!");
@@ -1194,7 +1190,6 @@ mod test {
                 selector: "destination.port".to_string(),
                 operator: WhenConditionOperator::Equal,
                 value: "8080".to_string(),
-                path: Default::default(),
                 compiled: Default::default(),
             };
             p.compile().expect("Should compile fine!");
@@ -1212,7 +1207,6 @@ mod test {
                 selector: "foobar".to_string(),
                 operator: WhenConditionOperator::Equal,
                 value: "8080".to_string(),
-                path: Default::default(),
                 compiled: Default::default(),
             };
             p.compile().expect("Should compile fine!");
@@ -1230,7 +1224,6 @@ mod test {
                 selector: "foobar".to_string(),
                 operator: WhenConditionOperator::Equal,
                 value: "1.0".to_string(),
-                path: Default::default(),
                 compiled: Default::default(),
             };
             p.compile().expect("Should compile fine!");
@@ -1248,7 +1241,6 @@ mod test {
                 selector: "connection.mtls".to_string(),
                 operator: WhenConditionOperator::Equal,
                 value: "true".to_string(),
-                path: Default::default(),
                 compiled: Default::default(),
             };
             p.compile().expect("Should compile fine!");
@@ -1266,7 +1258,6 @@ mod test {
                 selector: "request.time".to_string(),
                 operator: WhenConditionOperator::Equal,
                 value: "2023-05-28T00:00:00+00:00".to_string(),
-                path: Default::default(),
                 compiled: Default::default(),
             };
             p.compile().expect("Should compile fine!");
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -1,6 +1,6 @@
 use std::cell::OnceCell;
 use std::collections::HashMap;
-use std::fmt::{Debug, Display, Formatter};
+use std::fmt::{Debug, Formatter};
 use std::rc::Rc;
 use std::sync::Arc;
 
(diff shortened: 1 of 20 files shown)


adam-cattermole (commented):
Nice, the changes look good, just going to go through verification

src/configuration.rs
```diff
@@ -476,14 +389,6 @@ impl TryFrom<PluginConfiguration> for FilterConfig {
...
                     }
                 }
-                for action in &rule.actions {
-                    for datum in &action.data {
-                        let result = datum.item.compile();
-                        if result.is_err() {
-                            return Err(result.err().unwrap());
-                        }
-                    }
-                }
```
adam-cattermole, line 486: Actually one thing that's standing out to me is that we no longer keep the path as well as compiled version for PatternExpressions and don't pre-compile data items. I'm not 100%, does this conflict somewhat with the intention [here](https://github.com/Kuadrant/wasm-shim/issues/112#issuecomment-2419152322) cc @alexsnaps

alexsnaps: I need to look into those changes properly, but that would also mean that we'd err out at evaluation time instead of at config parse time... that seems... not desirable 🤔

eguzki: I removed because I did not see any point in early compilation. The "compilation" is just about building the path out of the selector string, which can never fail. Hence, I removed. 

Can you confirm my analysis? 

Anyway, I am happy to add that back.

alexsnaps: Doing it once, instead of on all invocations? Now we need to parse on every [`Policy::pattern_expression_applies`](https://github.com/Kuadrant/wasm-shim/pull/111/files#diff-016defdf85b98ea93c8767b2cbdce1d6eb7d8a1d0fe3c62a68e22e201c65c0efR61), right? And it will possibly fail when we move these to CEL expression as @adam-cattermole pointed out. But arguably that's my problem. 

In any case, I'm a big proponent of doing things as early as possible, represent them in the proper way from there on in the system and do the least amount of work. This code was early parsing the "path" and keeping it in proper form to avoid paying the computation time on the data plane's path. These are my guiding principle in general, but more so here.

eguzki: Thanks for confirming.

Bringing that back

eguzki:
Comments addressed. Ready for review @adam-cattermole @alexsnaps

adam-cattermole (approved):
Thanks for the changes, verification worked great

Issue #112, opened 2024-10-16
CEL expressions


eguzki:
Currently, *conditions*, *matches* (or whatever you want to call them), are expressed as a `PatternExpressions`

```rust
pub struct PatternExpression {                           
    pub selector: String,                                
    pub operator: WhenConditionOperator,                 
    pub value: String,                                   
                                                         
    #[serde(skip_deserializing)]                         
    path: OnceCell<Path>,                                
    #[serde(skip_deserializing)]                         
    compiled: OnceCell<CelExpression>,                   
}                                                        
```

Which in configuration looks like:

```yaml
matches:
    - selector: request.url_path
      operator: startswith
      value: /get
```

The ask here is to leverage the [Common Expression Language](https://github.com/google/cel-spec) to implement predicates that evaluate to `bool`.

Something like

```yaml
predicates:
    - "request.url_path.startsWith('/get')"
```

Note that data type on the configuration: CEL expressions will be strings that need to be parsed.

eguzki:
The [envoy attriubutes](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes) are available through a WASM-ABI, which in rust, and for Envoy, looks like 

```rust
pub fn get_property(path: Vec<&str>) -> Result<Option<Bytes>, Status>
```

There is no way to add all the attributes to the CEL context and make then available to expressions directly. So the question is about how to make _envoy_ attributes (and I underline Envoy here) on the CEL expressions. 

### Approach 1: CEL functions to the rescue

```yaml
predicates:
    - "get_property(request.url_path).startsWith('/get')"
```




> The Envoy locking here is undesirable, but that's what it is in the current state and context.

alexsnaps:
There is a third option, which is what actually happens today already:

parse the expression and replace member navigation in the AST directly. Today, when you have a condition that says `request.url_path`, [it gets flatten to a single binding: `attribute`](https://github.com/Kuadrant/wasm-shim/blob/main/src/configuration.rs#L364-L390). 

Also, there is more work that I think should happen as part of this: align authorino & limitador's lingo (or at least at the Kuadrant level). I was thinking having the so-called [`context: AttributeContext`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/attribute_context.proto) struct be reflected through all our CEL environments. wdyt?

alexsnaps:
[...]

PR #113, opened 2024-10-19
Predicate on our "well-known attributes"

alexsnaps:
fix to #112


alexsnaps:
This may need to supersede #110 ... stil wip tho

alexsnaps:
[tl;dr](https://github.com/Kuadrant/wasm-shim/pull/113/commits/5c558d4519a48bde4af8da179dcd9701d7e167d5#diff-f6344880d0def0b0b03411b5e9439442d8713a0447ad01d258ec087595dbc7e9R256-R260) this replaces `PatternExpression` with CEL `Predicates`, while keeping the former as a working fallback... 

Of interest if probably [this change](https://github.com/Kuadrant/wasm-shim/pull/113/commits/bf4f2f2754ab4460da86f5a7a59d627d9420c42b) as well, where the whole `fn` needs to be replace as `if cfg!(test)` within the method would work, but won't make the linker happy on anything but wasm...

eguzki:
I would change the PR tittle. That goes on the auto-generated release notes.

src/data/attribute.rs
```diff
@@ -129,7 +130,42 @@ pub fn store_metadata(metastruct: &Struct) {
     let metadata = process_metadata(metastruct, String::new());
     for (key, value) in metadata {
         let attr = format!("{KUADRANT_NAMESPACE}\\.{key}");
+        // stored into host_property: wasm.kuadrant.{key}
+        // example: wasm.kuadrant.identity.anonymous is how it's stored!
+        // but users would write the predicate: !auth.identity.anonymous
+        // two problems:
+        // - 1/ auth.identity doesn't resolve
+        // - 2/ the value is the string "true"
```
eguzki, line 138: >  the value is the string "true"

What is the problem  there? `auth.identity.anonymous` returns a string. We should document that and the user should deal with that. 

Besides, this something we can change from authorino, can't we? I mean, `auth.identity.anonymous` could be `bool` not string.

src/data/attribute.rs
```diff
@@ -129,7 +130,42 @@ pub fn store_metadata(metastruct: &Struct) {
...
+        //
+        // Admin can store this:
+        // authorino: export auth.identity.user.foo = expression("auth.user != null") // {"foo": true, "bar": 123, "name": "dd"}
+        // Or that:
+        // authorino: export auth.identity.foo = expression("auth.user.long.ass.path.to.some.member.within.foo")
+        // authorino: export auth.identity.user.bar = 123
+        // authorino: export auth.identity.user.name = "dd"
+
+        // predicate: !auth.identity.user.foo && auth.identity.user.bar != 443.0
+        //      => properties [["auth", "identity", "user", "foo"], ["destination", "port"]]
```
eguzki, line 154: where are those `destination` and `port` coming from?

src/data/attribute.rs
```diff
@@ -129,7 +130,42 @@ pub fn store_metadata(metastruct: &Struct) {
     let metadata = process_metadata(metastruct, String::new());
     for (key, value) in metadata {
         let attr = format!("{KUADRANT_NAMESPACE}\\.{key}");
+        // stored into host_property: wasm.kuadrant.{key}
```
eguzki, line 133: `wasm.kuadrant.{key}` or `kuadrant.{key}` ?

src/configuration.rs
```diff
@@ -169,6 +167,11 @@ impl PatternExpression {
             }
             Some(attribute_bytes) => attribute_bytes,
         };
+
+        // if someone would have the P_E be:
```
eguzki, line 171: I do not get this comment. For backwards compat?

why not :fire: to `PatternExpressions`?

src/configuration/action.rs
```diff
@@ -13,12 +14,24 @@ pub struct Action {
...
 }
 
 impl Action {
     pub fn conditions_apply(&self) -> bool {
-        self.conditions.is_empty() || self.conditions.iter().all(|m| m.applies())
+        let predicates = self
+            .compiled_predicates
+            .get()
+            .expect("predicates must be compiled by now");
+        if predicates.is_empty() {
```
eguzki, line 30: we need to document clearly that `predicates` take precedence and when at least one predicate is being defined, conditions are not even being evaluated.

eguzki (commented):
review still WIP

src/data/attribute.rs
alexsnaps: ... and json numbers are floats: `1` in JSON is `1_f64` using rust notation. It is not an integer.

src/data/attribute.rs
alexsnaps: `wasm.` prefix is automatically added by envoy. So you store `foobar`, but to retrieve it back you read `wasm.foobar`...

src/configuration.rs
alexsnaps: Yes! We should burn them as soon as the kuardrant operator doesn't use them to configure this wasm filter anymore...

src/configuration/action.rs
alexsnaps: We can... if this does NOT 🔥 by the release.

alexsnaps:
Just to be clear about the state of this... now that #110 was merged, while broken, let's not merge this quite yet. 
I took @didierofrivia work over in supporting "not so well known attribute", i.e. the data coming from authorino. I'll do that in 2 passes, scalar values (e.g. `auth.identity.anonymous` being resolved as being of type `bool` and then supporting any `expression` coming from authorino, expressed as JSON, in there too: List & Object)
I should get the former done by EOD EST, possibly the latter too...

src/configuration.rs
eguzki: yeah, that would allow us to merge this without waiting for operator to implement that.

alexsnaps:
[this change](https://github.com/Kuadrant/wasm-shim/pull/113/commits/f6ca3f6ecfb7492b7102aea744990cf6c12818b3) has the first step, but I think the latter step is _not_ required because of how the `struct` is decomposed in [process_metadata](https://github.com/Kuadrant/wasm-shim/blob/f8690551744a4edfd53b3b71fb5242b0cab8027b/src/data/attribute.rs#L173-L190).

src/data/cel.rs
```diff
@@ -0,0 +1,499 @@
...
+        let expression = parse(expression)?;
+
+        let mut props = Vec::with_capacity(5);
+        properties(&expression, &mut props, &mut Vec::default());
+
+        let attributes = props
+            .into_iter()
+            .map(|tokens| {
+                let path = Path::new(tokens);
+                known_attribute_for(&path).unwrap_or(Attribute {
```
eguzki, line 29: When a expression contains `request.headers.X-MY-HEADER`, that path is not a known attribute. Hence, the attribute is read expecting a String and returned bytes are parsed as String. 

However, the prefix of it, i.e. `request.headers`, is well known, a `map<string, string>`, how could we leverage that to be smarter?

eguzki (changes requested):
nice peace of work!

Some minor comments dropped.

Doc is required

src/data/cel.rs
alexsnaps: Yes... tho probably for now I would read the entire map and let the user deal as such with it. 
Another way is to have `properties` follow maps, as we know we can ask envoy to resolve these for us. 
Cause technically, that's the [way](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/main/src/hostcalls.rs#L140-L164)

src/data/cel.rs
```diff
@@ -0,0 +1,499 @@
...
+                let mut node = &mut root;
+                let mut it = attr.path.tokens().into_iter();
+                while let Some(token) = it.next() {
+                    if it.len() != 0 {
+                        node = match node
+                            .entry(token.to_string())
+                            .or_insert_with(|| Token::Node(HashMap::default()))
+                        {
+                            Token::Node(node) => node,
+                            Token::Value(_) => unreachable!(), // that's a bit of a lie!
```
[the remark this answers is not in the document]
alexsnaps: Yes, right now that is NOT supported. 
Ideally the 2nd entry would need to be filtered out. So I was thinking that one day we could sort, so that the shortest path are always first, and subsequent ones can "just" be ignored. 
It's not an issue right now, as you can't do anything with `auth` for instance right now really. Not entirely true, but say nothing useful. Also, that is also always better expressed without requiring the whole shorter path to be resolved. One exception is `request.headers` but that needs special attention anyways.

src/data/cel.rs
```diff
@@ -0,0 +1,499 @@
...
+
+    pub fn test(&self) -> bool {
+        match self.expression.eval() {
+            Value::Bool(result) => result,
+            _ => false,
+        }
+    }
+}
+
+pub struct Attribute {
```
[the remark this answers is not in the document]
alexsnaps: Well yes, but decided _not_ because it has the concept of _known attribute types_ that are CEL. So it could actually be `CelTypedAttribute` instead...

src/data/cel.rs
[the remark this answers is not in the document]
eguzki: Have you considered building another type of tree structure?

```rust
struct Node {
     value: Option<Attribute>,  
     children: Vec<Node>,
} 

struct Tree {
     roots: Vec<Node>,
}
```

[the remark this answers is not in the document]
alexsnaps: To what purpose? I assume you're talking about this input:

```
[
   Attribute(["auth"]),
   Attribute(["auth", "identity")
]
```
yes? If so, let's zoom back for a second. if you access `auth`, I need to hydrate `auth` in full... So that `["auth", "identity"]` is of no use anymore, at the `identity` member was hydrated as part of `auth`, right? 
Or am I missing the use case you are trying to address here?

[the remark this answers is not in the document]
alexsnaps: I've ordered and skip an "more precise" attribute when a Value was resolved earlier in the graph.
So if `auth` was resolved, no need to resolve `auth.identity`, as its a member (hopefully) of `auth` that's now fully hydrated, see [here](https://github.com/Kuadrant/wasm-shim/pull/113/commits/dbdf7e10a7f3f9286143d3ad4ad655255f70e9fc)

src/data/cel.rs
alexsnaps: Also, this `request.headers.X-MY-HEADER` is not a valid expression, it should be `request.headers["X-MY-HEADER"]`

src/data/cel.rs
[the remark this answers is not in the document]
eguzki: I was thinking on 

```
contains(request.headers, "X-MY-HEADER") && request.headers.OTHER-HEADER == "some-value"
```

But, no need to address this kind of issues now. Nitpicking here.

alexsnaps:
Alright, it should "all" work. i.e. now we store JSON representation of the `auth` data coming back from authorino, so to not lose the type information... We even get integers! I also added support for getting to headers...

src/data/property.rs
```diff
@@ -37,6 +38,33 @@ fn host_get_property(path: &Path) -> Result<Option<Vec<u8>>, Status> {
...
 
+#[cfg(test)]
+pub fn host_get_map(path: &Path) -> Result<HashMap<String, String>, String> {
+    match *path.tokens() {
+        ["request", "headers"] => Ok(HashMap::default()),
+        _ => Err(format!("Unknown map requested {:?}", path)),
+    }
+}
+
+#[cfg(not(test))]
```
didierofrivia, line 49: I like this way of defining the fn when it's meant to be mocked for testing and not. Might be worthy to follow the same pattern for the operation dispatcher and grpc message req/res //TODO:(didierofrivia)

didierofrivia (approved):
LGTM. 💪🏼 🥇 🙏🏼

src/data/cel.rs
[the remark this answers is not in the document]
alexsnaps: Which btw is why it's in the `cel` ~file~ module.

src/configuration/action.rs
```diff
@@ -47,13 +49,28 @@ impl Action {
...
+                    (static_item.key.to_owned(), static_item.value.to_owned())
                 }
+                DataType::Expression(cel) => (
+                    cel.key.clone(),
+                    match cel
+                        .compiled
+                        .get()
+                        .expect("Expression must be compiled by now")
+                        .eval()
+                    {
```
alexsnaps, line 63: This is how we've send data to limitador so far, losing the type information... just a string: `true`, `42`, `bob`

src/configuration/action.rs
```diff
@@ -47,13 +49,28 @@ impl Action {
...
+                        Value::Int(n) => format!("{n}"),
+                        Value::UInt(n) => format!("{n}"),
+                        Value::Float(n) => format!("{n}"),
+                        // todo this probably should be a proper string literal!
+                        Value::String(s) => (*s).clone(),
+                        Value::Bool(b) => format!("{b}"),
+                        Value::Null => "null".to_owned(),
```
alexsnaps, line 70: Wondering if we should skip that descriptor entry... this _could_ happen if a value is explicitly set to `null` in `auth` ...

alexsnaps:
Status update: this 🔥 the need for anything but CEL. While the previous ways are still there, they should _NOT_ be used anymore and comes with bunch of caveat... I'll let the code speak for itself as how one can shoot themselves in the foot by going to raw properties/attribute values.

src/data/attribute.rs
```diff
@@ -128,7 +129,7 @@ pub fn set_attribute(attr: &str, value: &[u8]) {
 pub fn store_metadata(metastruct: &Struct) {
     let metadata = process_metadata(metastruct, String::new());
     for (key, value) in metadata {
-        let attr = format!("{KUADRANT_NAMESPACE}\\.{key}");
+        let attr = format!("{KUADRANT_NAMESPACE}\\.auth\\.{key}");
```
eguzki, line 132: I was thinking that the `auth` prefix should come from `pub fn process_auth_grpc_response` method.

Then, `pub fn store_metadata(metastruct: &Struct)` could be `pub fn store_metadata(prefix: &str, metastruct: &Struct)`

My diff

```diff
diff --git a/src/data/attribute.rs b/src/data/attribute.rs
index 9e65caa..7888462 100644
--- a/src/data/attribute.rs
+++ b/src/data/attribute.rs
@@ -126,10 +126,10 @@ pub fn set_attribute(attr: &str, value: &[u8]) {
     };
 }
 
-pub fn store_metadata(metastruct: &Struct) {
+pub fn store_metadata(prefix: &str, metastruct: &Struct) {
     let metadata = process_metadata(metastruct, String::new());
     for (key, value) in metadata {
-        let attr = format!("{KUADRANT_NAMESPACE}\\.auth\\.{key}");
+        let attr = format!("{KUADRANT_NAMESPACE}\\.{prefix}\\.{key}");
         debug!("set_attribute: {attr} = {value}");
         set_attribute(attr.as_str(), value.into_bytes().as_slice());
     }
diff --git a/src/service/auth.rs b/src/service/auth.rs
index c8e4a4d..8364a8b 100644
--- a/src/service/auth.rs
+++ b/src/service/auth.rs
@@ -128,7 +128,7 @@ impl AuthService {
     ) -> Result<(), StatusCode> {
         if let GrpcMessageResponse::Auth(check_response) = auth_resp {
             // store dynamic metadata in filter state
-            store_metadata(check_response.get_dynamic_metadata());
+            store_metadata("auth", check_response.get_dynamic_metadata());
 
             match check_response.http_response {
                 Some(CheckResponse_oneof_http_response::ok_response(ok_response)) => {
```

src/data/attribute.rs
```diff
@@ -128,7 +129,7 @@ pub fn set_attribute(attr: &str, value: &[u8]) {
...
         debug!("set_attribute: {attr} = {value}");
```
eguzki, line 133: What about this `debug!` macro in `pub fn set_attribute(attr: &str, value: &[u8])` ?

src/data/attribute.rs
eguzki, line 132: Can you please add a comment saying that attributes set here are accessible with `wasm.*` prefix?

src/configuration.rs
```diff
@@ -22,6 +23,23 @@ pub mod action;
 pub mod action_set;
 mod action_set_index;
 
+#[derive(Deserialize, Debug, Clone)]
+pub struct ExpressionItem {
+    pub key: String,
```
eguzki, line 28: can we make `key` optional? it would default to the expression value (not the referenced value, though).

eguzki (commented):
Left some comments and food for thoughts. 

This is not about requesting changes. 

From my side, this is ready to be merged

src/configuration.rs
alexsnaps: I think it makes for a very horrible default, but... 🤷

alexsnaps:
@eguzki This is my gift to you... you all of course :)
Do whatever you want with the PR! You have my blessing. I need to move on here...

eguzki (approved):

PR #114, opened 2024-10-22
dev environment: fix authorino deployment

eguzki:
### What

Authorino deployment changed and webhook is no longer needed https://github.com/Kuadrant/authorino-operator/pull/219

* Removed TLS overlay, as not relevant for the wasm-shim.
* Leverage limtiador operator for deployment.


didierofrivia (approved):
🧑‍🏭

eguzki:
@adam-cattermole merging this. 

Please let me know if TLS overlay is still useful for the dev environment. I am more than happy to bring it back. I just do not see a value of the TLS which is totally delegated to Envoy and we were duplicating wasm configurations in both overlies.

PR #116, opened 2024-10-23
Updating Dev testing env

didierofrivia:
This PR updates our development environment for testing and the e2e test for the remote address using CEL expressions and predicates.

### DIFF
--- a/utils/deploy/envoy.yaml
+++ b/utils/deploy/envoy.yaml
@@ -337,7 +337,7 @@ data:
                                           "data": [
                                             {
                                               "selector": {
-                                                "selector": "filter_state.wasm\\.kuadrant\\.identity\\.userid",
+                                                "selector": "auth.identity.userid",
                                                 "key": "user_id"
                                               }
                                             }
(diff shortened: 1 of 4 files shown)


utils/deploy/envoy.yaml
```diff
@@ -337,7 +337,7 @@ data:
                                           "data": [
                                             {
                                               "selector": {
-                                                "selector": "filter_state.wasm\\.kuadrant\\.identity\\.userid",
+                                                "selector": "auth.identity.userid",
```
alexsnaps, line 340: This is probably not something one would ever want to do. 
All `selector` usages should be avoided. This should be an `expression` here.

utils/deploy/envoy.yaml
```diff
@@ -212,7 +204,7 @@ data:
                                           "scope": "rlp-ns-B/rlp-name-B",
                                           "data": [
                                             {
-                                              "static": {
+                                              "expression": {
                                                 "key": "rlp-ns-B/rlp-name-B/limit-not-to-be-activated",
                                                 "value": "1"
```
alexsnaps, line 209: While this would work, this also changed the type tho! It's now `int(1)` ... but will be erased and will be `1`, just as if you had it as `"\"1\""` in there, which would be closer semantically to what this was before...

tl;dr _not wrong_, but... ;)

eguzki (approved):

Issue #123, opened 2024-10-28
missing cluster error handling


eguzki:
When the service endpoint does not exist from the defined set of envoy clusters.

```json
"services": {
  "mistyped-service": {
    "type": "ratelimit",
    "endpoint": "does-not-exist",
    "failureMode": "deny"
  }
},
"actionSets": [
{
    "actions": [
        {
            "service": "mistyped-service",
            "scope": "b",
            "data": [
                {
                    "expression": {
                        "key": "limit_to_be_activated",
                        "value": "1"
                    }
                }
            ]
        }
    ]
}
]
```

Then, the request will `Continue` regardless of `failureMode` config. 

I have implemented an e2e test that reproduces this issue: https://github.com/Kuadrant/wasm-shim/pull/122

didierofrivia:
I'm a bit confused, since https://github.com/Kuadrant/wasm-shim/pull/122 is about parallel requests to envoy and this issue is about error handling? The implementation of the operation dispatcher is a basic one, I agree we need error handling (among other things), but mistyped configuration would kind of be a thing coming from the Kuadrant Operator.
In any case, if a given request to a service fails, we should follow the `failureMode` config for that service and Allow/Deny respectively **or** define a set of operations to follow depending on the service? (i.e: retry, authorize, limit...etc)

eguzki:
> In any case, if a given request to a service fails, we should follow the failureMode config for that service and Allow/Deny respectively or define a set of operations to follow depending on the service? 

I would say follow the `failureMode`. :shrug:

PR #125, opened 2024-10-29
Operation dispatcher err handling

didierofrivia:
Fixes https://github.com/Kuadrant/wasm-shim/issues/123

Also fixes the following e2e tests:
* https://github.com/Kuadrant/wasm-shim/pull/124
* https://github.com/Kuadrant/wasm-shim/pull/121
* https://github.com/Kuadrant/wasm-shim/pull/122

This PR introduces a refactor that besides fixing the issue #123, makes the code a bit more consistent regarding the flow of the `Operation`s State controlled by the `OperationHandler`. It aims to handle all the exceptions raised and report back to the Filter (or any caller).

### DIFF
--- a/src/filter/http_context.rs
+++ b/src/filter/http_context.rs
@@ -4,7 +4,7 @@ use crate::operation_dispatcher::{OperationDispatcher, OperationError};
 use crate::service::GrpcService;
 use log::{debug, warn};
 use proxy_wasm::traits::{Context, HttpContext};
-use proxy_wasm::types::{Action, Status};
+use proxy_wasm::types::Action;
 use std::cell::RefCell;
 use std::rc::Rc;
 
@@ -53,37 +53,32 @@ impl Filter {
             );
             return Action::Continue;
         }
-        let res = self.operation_dispatcher.borrow_mut().next();
-        if let Ok(operation) = res {
-            match operation.get_result() {
+
+        match self.operation_dispatcher.borrow_mut().next() {
+            Ok(Some(op)) => match op.get_result() {
                 Ok(call_id) => {
                     debug!("#{} initiated gRPC call (id# {})", self.context_id, call_id);
                     Action::Pause
                 }
                 Err(e) => {
                     warn!("gRPC call failed! {e:?}");
-                    if let FailureMode::Deny = operation.get_failure_mode() {
+                    if let FailureMode::Deny = op.get_failure_mode() {
                         self.send_http_response(500, vec![], Some(b"Internal Server Error.\n"))
                     }
                     Action::Continue
                 }
+            },
+            Ok(None) => {
+                Action::Continue // No operations left to perform
             }
-        } else {
-            let op_err = res.unwrap_err();
-            match op_err {
-                OperationError {
-                    status: Status::Empty,
-                    ..
-                } => Action::Continue,
-                OperationError {
-                    failure_mode: FailureMode::Deny,
-                    ..
-                } => {
-                    self.send_http_response(500, vec![], Some(format!("{op_err}").as_ref()));
-                    Action::Continue
-                }
-                _ => Action::Continue,
+            Err(OperationError {
+                failure_mode: FailureMode::Deny,
+                ..
+            }) => {
+                self.send_http_response(500, vec![], Some(b"Internal Server Error.\n"));
+                Action::Continue
             }
+            _ => Action::Continue,
         }
     }
 }
@@ -136,7 +131,7 @@ impl Context for Filter {
         match op_res {
             Ok(operation) => {
                 if GrpcService::process_grpc_response(operation, resp_size).is_ok() {
-                    if let Ok(_op) = self.operation_dispatcher.borrow_mut().next() {
+                    if let Ok(Some(_op)) = self.operation_dispatcher.borrow_mut().next() {
                     } else {
                         self.resume_http_request()
                     }
--- a/src/filter/http_context.rs
+++ b/src/filter/http_context.rs
@@ -1,6 +1,6 @@
 use crate::configuration::action_set::ActionSet;
 use crate::configuration::{FailureMode, FilterConfig};
-use crate::operation_dispatcher::OperationDispatcher;
+use crate::operation_dispatcher::{OperationDispatcher, OperationError};
 use crate::service::GrpcService;
 use log::{debug, warn};
 use proxy_wasm::traits::{Context, HttpContext};
@@ -65,12 +65,16 @@ impl Filter {
                 }
             }
         } else {
-            match res {
-                Err((Status::Empty, _)) => Action::Continue,
-                Err((_, failure_mode)) => {
-                    if failure_mode == FailureMode::Deny {
-                        self.send_http_response(500, vec![], Some(b"Internal Server Error.\n"))
-                    }
+            match res.unwrap_err() {
+                OperationError {
+                    status: Status::Empty,
+                    ..
+                } => Action::Continue,
+                OperationError {
+                    failure_mode: FailureMode::Deny,
+                    ..
+                } => {
+                    self.send_http_response(500, vec![], Some(b"Internal Server Error.\n"));
                     Action::Continue
                 }
                 _ => Action::Continue,
@@ -131,7 +135,7 @@ impl Context for Filter {
             }
         } else {
             warn!("No Operation found with token_id: {token_id}");
-            GrpcService::handle_error_on_grpc_response(&FailureMode::Deny); // TODO(didierofrivia): Decide on what's the default failure mode
+            GrpcService::handle_error_on_grpc_response(FailureMode::Deny); // TODO(didierofrivia): Decide on what's the default failure mode
         }
     }
 }
--- a/src/filter/http_context.rs
+++ b/src/filter/http_context.rs
@@ -131,10 +131,15 @@ impl Context for Filter {
         match op_res {
             Ok(operation) => {
                 if GrpcService::process_grpc_response(operation, resp_size).is_ok() {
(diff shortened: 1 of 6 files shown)


src/filter/http_context.rs
```diff
@@ -114,15 +127,15 @@ impl Context for Filter {
```
eguzki, line 127: I think we should check `status_code`. When grpc endpoint is unreachable, this is flagged on the `status_code`.

src/operation_dispatcher.rs
```diff
@@ -162,14 +195,14 @@ impl OperationDispatcher {
...
+                                        Ok(operation.clone())
                                     }
                                     State::Done => self.next(),
                                 }
                             }
-                            Err(status) => {
-                                error!("{status:?}");
-                                None
+                            Err(err) => {
+                                error!("{err:?}");
```
eguzki, line 204: I wonder if we need to log here? error is being returned

src/filter/http_context.rs
```diff
@@ -114,15 +127,15 @@ impl Context for Filter {
 
         if let Some(operation) = some_op {
             if GrpcService::process_grpc_response(operation, resp_size).is_ok() {
-                self.operation_dispatcher.borrow_mut().next();
-                if let Some(_op) = self.operation_dispatcher.borrow_mut().next() {
+                let _ = self.operation_dispatcher.borrow_mut().next();
```
eguzki, line 130: this operation can return error, shouldn't be handled?

src/filter/http_context.rs
```diff
@@ -114,15 +127,15 @@ impl Context for Filter {
...
+                if let Ok(_op) = self.operation_dispatcher.borrow_mut().next() {
```
eguzki, line 131: Error handling?

src/operation_dispatcher.rs
```diff
@@ -191,7 +224,7 @@ impl OperationDispatcher {
                 }
             }
         } else {
-            None
+            Err(OperationError::new(Status::Empty, FailureMode::default())) // No more operations
```
eguzki, line 227: no more op is an error? maybe return 

```rust
Result<Option<Rc<Operation>>, OperationError>
```

eguzki (commented):
few comments dropped

src/filter/http_context.rs
eguzki: I do not understand why we need  to call twice the operation distpatcher

eguzki:
This is still not working for me. I run the following configuration:

```json

{
  "services": {
    "limitadorA": {
      "type": "ratelimit",
      "endpoint": "limitador",
      "failureMode": "deny"
    },
    "limitadorB": {
      "type": "ratelimit",
      "endpoint": "unknown",
      "failureMode": "deny"
    }
  },
  "actionSets": [
    {
      "name": "basic",
      "routeRuleConditions": {
        "hostnames": [
          "*.example.com"
        ]
      },
      "actions": [
        {
          "service": "limitadorA",
          "scope": "basic",
          "data": [
            {
              "expression": {
                "key": "a",
                "value": "1"
              }
            }
          ]
        },
        {
          "service": "limitadorB",
          "scope": "basic",
          "data": [
            {
              "expression": {
                "key": "a",
                "value": "1"
              }
            }
          ]
        }
      ]
    }
  ]
}
```
There are  two actions, both with `failureMode` to `false`. And the second one targets a missing cluster. On running a request, the request makes its way to upstream and returns 200 regardless of `failureMode`.

didierofrivia:
Now @eguzki ! this is your time to shine !!! (and thanks in advance for the reviews 🙏🏼 )

didierofrivia:
(I've included the commits from your test, just for easy testing... I could remove them from this PR if you want)

src/filter/http_context.rs
```diff
@@ -50,22 +54,31 @@ impl Filter {
...
-        } else {
-            Action::Continue
+            Err(OperationError {
+                failure_mode: FailureMode::Deny,
+                ..
+            }) => {
+                self.send_http_response(500, vec![], Some(b"Internal Server Error.\n"));
+                Action::Continue
+            }
+            _ => Action::Continue,
```
didierofrivia, line 81: we could omit 

```rust
Ok(None) => {
                Action::Continue // No operations left to perform
            }
```

Left it there just so we now explicitly in case we want to perform something else

src/operation_dispatcher.rs
```diff
@@ -92,20 +98,47 @@ impl Operation {
...
+        Self {
+            status,
+            failure_mode,
+        }
+    }
+}
+
+impl fmt::Display for OperationError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self.status {
```
didierofrivia, line 134: Depending on the status we could add specific messages

src/operation_dispatcher.rs
didierofrivia: we could, we were using it for debugging. not strong opinion about keeping it

didierofrivia:
By the way @eguzki, this PR also fixes your https://github.com/Kuadrant/wasm-shim/pull/122/ ;)

src/filter/http_context.rs
```diff
@@ -110,19 +123,29 @@ impl Context for Filter {
...
-                    self.resume_http_request()
+        match op_res {
+            Ok(operation) => {
+                if GrpcService::process_grpc_response(operation, resp_size).is_ok() {
+                    // call the next op
+                    match self.operation_dispatcher.borrow_mut().next() {
+                        Ok(_) => {} // no action needed
+                        Err(op_err) => {
+                            // If desired, we could check the error status.
+                            GrpcService::handle_error_on_grpc_response(op_err.failure_mode);
```
eguzki, line 139: Here we handle error and then we resume http request with ` self.resume_http_request();`

I wonder if that is correct.

src/filter/http_context.rs
didierofrivia: it might be useful, however I'm a bit confused about what are the `status_code` valid values... since there's no docs and it's confusing their examples... would it be a [GrpcStatusCode](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/main/src/types.rs#L121) or [Status](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/main/src/types.rs#L43).. then it's a bit confusing in this [particular example](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/main/examples/grpc_auth_random/src/lib.rs#L70-L80) how they treat it. What do you reckon? what meaningful check can we do to avoid parsing and returning early?

src/filter/http_context.rs
didierofrivia: Actually, the flow should be:

1. On error:
  a. FailureMode::Deny => return 500
  b FailureMode::Allow => resume request
  
2. On op => nothing, already called the next op
3. On None => No more ops, resume request

will push another commit. Thanks!

src/operation_dispatcher.rs
didierofrivia: let me know and I can remove it

eguzki: leave it, not a big issue.

src/filter/http_context.rs
eguzki: From the tests I have run, on successful GRPC request/response exchange, `status_code` is [0](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/main/src/types.rs#L122). Whereas when the service is unreachable, `status_code` is [14](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/main/src/types.rs#L136).

That all I know for now.

Let's merge this and you can enhance this in a following up PR's

eguzki (approved):
working software

All tests passing:
* https://github.com/Kuadrant/wasm-shim/pull/124
* https://github.com/Kuadrant/wasm-shim/pull/122
* https://github.com/Kuadrant/wasm-shim/pull/121
* https://github.com/Kuadrant/wasm-shim/pull/128

plus existing unittests

src/filter/http_context.rs
alexsnaps: So, for all _other_ `Err`, we ... `Action::Continue`?

src/operation_dispatcher.rs
alexsnaps: Right, this is... a little random. Logging at _error creation_ time, possibly at a _lower_ level, e.g. `debug`, is something I'd get. Here,... what's the point? In the middle of the call stack we grab the error and log it 🤔 

Might be because you had to deal with it to "only" rewrap it? Cause you could actually get away with _not_ doing that with the [`?`](https://doc.rust-lang.org/rust-by-example/std/result/question_mark.html)

alexsnaps (commented):
Thinking error handling from performing an operation could be pushed within the Operations themselves, making the orchestration of the pipeline/operation chain a little easier on the reader... not asking to change this tho.

adam-cattermole (approved):

src/filter/http_context.rs
alexsnaps: I guess I'm saying that I'd leave the `Ok(None)` => Action::Continue,` as an explicit branch _but_ would have 3 `Err` branches: 

 - `FailureMode::Deny`, as you do (log the actual error there? as this where it all ends, right?)
 - `FailureMode::Allow`, explicit with the actual error logged again
 - `Err(_)`, that possibly `panic!` or is marked as `unreachable!` as that I think is the assumption this builds upon

didierofrivia: I made it explicit the handling of both failure modes, but the third branch makes clippy unhappy. I'm also logging the status. Would that be enough?

didierofrivia: Let me know and I can include it in a following PR ;)

PR #126, opened 2024-10-29
Support request.query decoding

alexsnaps:
See test, but this effectively works like this, given the query string `param1=%F0%9F%91%BE%20&param2=Exterminate%21&%F0%9F%91%BE=123&%F0%9F%91%BE=456`

```
decodeQueryString(request.query)['param1'] == '👾 ' &&
decodeQueryString(request.query)['param2'] == 'Exterminate!' &&
decodeQueryString(request.query)['👾'][0] == '123' &&
decodeQueryString(request.query)['👾'][1] == '456'
```

i.e.:
 - Doesn't trim
 - decodes all utf-8
 - supports lists of values

It is exposed as an "extension" that's only enabled via `Expression::new_extended`, but mostly thru `Predicate::route_rule` that's now _only_ used in the config, when compiling `RouteRuleConditions`

### DIFF
--- a/src/data/cel.rs
+++ b/src/data/cel.rs
@@ -1,6 +1,6 @@
 use crate::data::get_attribute;
 use crate::data::property::{host_get_map, Path};
-use cel_interpreter::extractors::This;
+use cel_interpreter::extractors::{Arguments, This};
 use cel_interpreter::objects::{Key, Map, ValueType};
 use cel_interpreter::{Context, ExecutionError, ResolveResult, Value};
 use cel_parser::{parse, Expression as CelExpression, Member, ParseError};
@@ -85,8 +85,17 @@ impl Expression {
 
 /// Decodes the query string and returns a Map where the key is the parameter's name and
 /// the value is either a [`Value::String`] or a [`Value::List`] if the parameter's name is repeated
+/// and the second arg is set not set to `false`.
 /// see [`tests::decodes_query_string`]
-fn decode_query_string(This(s): This<Arc<String>>) -> ResolveResult {
+fn decode_query_string(This(s): This<Arc<String>>, Arguments(args): Arguments) -> ResolveResult {
+    let allow_repeats = if args.len() == 2 {
+        match &args[1] {
+            Value::Bool(b) => *b,
+            _ => true,
+        }
+    } else {
+        true
+    };
     let mut map: HashMap<Key, Value> = HashMap::default();
     for part in s.split('&') {
         let mut kv = part.split('=');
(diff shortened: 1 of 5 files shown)


eguzki (changes requested):
LGTM

Add some doc and ready to be merged

src/data/cel.rs
```diff
@@ -63,6 +67,27 @@ impl Expression {
...
 }
 
+fn url_decode(This(s): This<Arc<String>>) -> ResolveResult {
+    let mut map: HashMap<Key, Value> = HashMap::default();
+    for part in s.split('&') {
+        let mut kv = part.split('=');
+        if let (Some(key), Some(value)) = (kv.next(), kv.next()) {
+            match map.entry(decode(key).unwrap().into_owned().into()) {
+                Entry::Occupied(mut e) => {
+                    let v = e.get().clone();
```
alexsnaps, line 77: Need to check whether `v` is a `Value::List` here. Add to it if it is, otherwise `clone()` and replace the list as we do here.

src/data/cel.rs
```diff
@@ -492,6 +521,37 @@ mod tests {
...
+        .expect("This is valid!");
+        assert!(predicate.test());
+
+        super::super::property::test::TEST_PROPERTY_VALUE.set(Some(
+            "%F0%9F%91%BE"
+                .bytes()
+                .collect(),
+        ));
+        let predicate = Predicate::new(
+            "urlDecode(request.query) == {'👾': ''}
```
alexsnaps, line 548: Should this return a `Value::String` ?

alexsnaps:
@guicassolato so this works for your use-case(s)?
Also, that [last test](https://github.com/Kuadrant/wasm-shim/pull/126/files#r1822212963) is somewhat sad... wdy'all think?

didierofrivia (approved):
👾

alexsnaps:
So reading this: 

```
// If a query param is repeated in an HTTP request, the behavior is
// purposely left undefined, since different data planes have different
// capabilities. However, it is *recommended* that implementations should
// match against the first value of the param if the data plane supports it,
// as this behavior is expected in other load balancing contexts outside of
// the Gateway API.
//
// Users SHOULD NOT route traffic based on repeated query params to guard
// themselves against potential differences in the implementations.
```

Wondering if we shouldn't always have the first param in our `Map` and drop the `List` altogether... 
I have a PR getting ready that would: 

```go
fmt.Sprintf("decodeQueryString(request.query)['%s'] == '%s'", queryParam.Name, queryParam.Value)
``` 

@guicassolato do you have an opinion?

alexsnaps:
Alternate proposal, that I'll implement here right now, pass an optional 2nd bool param, that would dictate whether to return a `List` when repeated params are present, or only the first one (`includeRepeatedParams: bool` kinda thing).

eguzki:
In description it says `decodeQueryString` but in the implementation I see `queryMap`.

eguzki (approved):

PR #129, opened 2024-11-02
Assert on the path requested when accessing well known attr

alexsnaps:
Makes sure you are actually accessing the path you expect... 
Most useful to document [these](https://github.com/Kuadrant/wasm-shim/pull/129/files#diff-030f3ad71048fac9886241d6c33bcd3741a027e0403b506e115f5a6c5e37dcb7R477-R482)

### DIFF
--- a/src/data/cel.rs
+++ b/src/data/cel.rs
@@ -449,12 +449,14 @@ pub mod data {
 #[cfg(test)]
 mod tests {
     use crate::data::cel::{known_attribute_for, Expression, Predicate};
+    use crate::data::property;
     use cel_interpreter::objects::ValueType;
 
     #[test]
     fn predicates() {
         let predicate = Predicate::new("source.port == 65432").expect("This is valid CEL!");
-        super::super::property::test::TEST_PROPERTY_VALUE.set(Some(65432_i64.to_le_bytes().into()));
+        property::test::TEST_PROPERTY_VALUE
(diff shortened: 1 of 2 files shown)


src/data/cel.rs
```diff
@@ -470,36 +472,65 @@ mod tests {
...
         let value = Expression::new("auth.identity.age").unwrap().eval();
+        assert_eq!(value, "John".into());
+
+        property::test::TEST_PROPERTY_VALUE.set(Some((
+            property::Path::new(vec!["filter_state", "wasm.kuadrant.auth.identity.name"]),
+            "-42".bytes().collect(),
+        )));
+        let value = Expression::new("auth.identity.name").unwrap().eval();
         assert_eq!(value, (-42).into());
+
```
adam-cattermole, line 512: I guess it doesn't really matter as such as the path is somewhat irrelevant to the change but minor nit; age == John, name == -42

adam-cattermole (approved):
lgtm

src/data/cel.rs
alexsnaps: I actually did it on purpose, to _not_ imply typing based of what otherwise could

adam-cattermole: Ah, sure!

didierofrivia (approved):
🧑‍🚀

PR #132, opened 2024-11-04
CEL Strings extension macros

alexsnaps:


src/data/cel/strings.rs
```diff
@@ -0,0 +1,374 @@
...
+                        ),
+                    })
+                }
+            };
+            if base >= this.len() {
+                return Ok((-1).into());
+            }
+            match this[base..].rfind(&*arg) {
+                None => Ok((-1).into()),
+                Some(idx) => Ok(Value::UInt(idx as u64)),
```
adam-cattermole, line 83: Shouldn't this be `base +` similar to `index_of`?
```suggestion
                Some(idx) => Ok(Value::UInt(base + idx as u64)),
```

alexsnaps: It doesn't try... and tests won't pass (the tests are from the doc et al from the golang version)
That being said, when I read the expectations for that method, I failed to understand them

src/data/cel/strings.rs
```diff
@@ -0,0 +1,374 @@
...
+                .take_while(|_| {
+                    if let Some(v) = end.checked_sub(1) {
+                        end = v;
+                        true
+                    } else {
+                        false
+                    }
+                })
+                .collect::<String>()
+                .into())
```
adam-cattermole, line 269: Maybe I'm missing a case but couldn't we just avoid this `take_while`? although maybe this is related to what we

```suggestion
            Ok(this
                .chars()
                .skip(start)
                .take(end - start)
                .collect::<String>()
                .into())
```

src/data/cel/strings.rs
adam-cattermole: Sure, thanks for [this link](https://github.com/google/cel-go/blob/master/ext/strings.go), I see this is expected

adam-cattermole (approved):
Has a merge conflict that needs resolving but the rest lgtm

PR #136, opened 2024-11-07
doc new features

eguzki:
* String extension funcs
* Query string decoding `queryMap`
* `getHostProperty`

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,92 @@ evaluating to a `bool` value, while `Expression`, used for passing data to a ser
 
 These expression can operate on the data made available to them through the Well Known Attributes, see below
 
+#### CEL String Extended Functions
+
+In addition to the existing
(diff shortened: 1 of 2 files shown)


alexsnaps:
I really don't understand this! This is "shoveling snow forward" as we say in French Canada: 
>[pelleter par en avant](https://ferd.ca/on-technical-debt-shoveling-forward.html)

We're literally creating ~technical~ documentation debt for ourselves in my opinion. So my questions are:

 - Who are you writing that documentation for?
 - What do you think you're achieving by adding this? And doing this in _this very way_?
 
I have opinion on both of these, but I am _really trying_ to understand where _you_ are coming from?

eguzki:
Documenting exposed functions and capabilities for anyone writing wasm configuration and CEL predicates

alexsnaps:
The best I could find wrt the `Strings` extension is from the [`README.md`](https://github.com/google/cel-go/tree/master/ext#strings) on the cel-go implementation

PR #137, opened 2024-11-08
remove unused conditions

eguzki:
Kuadrant operator no longer uses conditions based on `PatternExpressions`:  https://github.com/Kuadrant/kuadrant-operator/pull/988

This PR removes conditions based on `PatternExpressions` on both `RouteRuleConditions` and `Action`

This PR removes `Selector` *DataType*


tests/multi.rs
```diff
@@ -699,11 +682,11 @@ fn authenticated_one_ratelimit_action_matches() {
                 10, 215, 1, 10, 20, 10, 18, 10, 16, 18, 10, 49, 46, 50, 46, 51, 46, 52, 58, 56, 48,
                 24, 200, 223, 2, 18, 23, 10, 21, 10, 19, 18, 14, 49, 50, 55, 46, 48, 46, 48, 46,
                 49, 58, 56, 48, 48, 48, 24, 192, 62, 34, 141, 1, 10, 0, 18, 136, 1, 18, 3, 71, 69,
-                84, 26, 30, 10, 10, 58, 97, 117, 116, 104, 111, 114, 105, 116, 121, 18, 16, 97, 98,
-                105, 95, 116, 101, 115, 116, 95, 104, 97, 114, 110, 101, 115, 115, 26, 14, 10, 7,
-                58, 109, 101, 116, 104, 111, 100, 18, 3, 71, 69, 84, 26, 38, 10, 5, 58, 112, 97,
-                116, 104, 18, 29, 47, 100, 101, 102, 97, 117, 108, 116, 47, 114, 101, 113, 117,
-                101, 115, 116, 47, 104, 101, 97, 100, 101, 114, 115, 47, 112, 97, 116, 104, 34, 10,
+                84, 26, 14, 10, 7, 58, 109, 101, 116, 104, 111, 100, 18, 3, 71, 69, 84, 26, 38, 10,
```
eguzki, line 685: @adam-cattermole I cannot explain the reasoning for this change needed (otherwise test will fail). Any hint?

src/configuration.rs
```diff
@@ -7,14 +7,11 @@ use std::sync::Arc;
 use crate::configuration::action_set::ActionSet;
 use crate::configuration::action_set_index::ActionSetIndex;
 use crate::data;
+use crate::data::Predicate;
 use crate::data::PropertyPath;
```
alexsnaps, line 11: Can't this be removed too and get us the same benefits (with the additional same changes to `data/mod.rs` and making the struct `pub(super)`) as `get_property`?

src/configuration/action_set.rs
```diff
@@ -44,15 +41,8 @@ impl ActionSet {
...
-        if predicates.is_empty() {
-            self.route_rule_conditions.matches.is_empty()
-                || self
-                    .route_rule_conditions
-                    .matches
-                    .iter()
-                    .all(|m| m.applies())
-        } else {
-            predicates
+        predicates.is_empty()
```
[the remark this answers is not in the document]
eguzki: Yeah, I realized that and working on that and adding tests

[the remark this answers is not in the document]
eguzki: done

tests/multi.rs
adam-cattermole: Hmm I saw something similar when looking at this recently, when I debug printed the check request I could see the order of the headers was sometimes changing - unsure exactly on the cause or if that's what you're seeing here without printing the actual struct

src/configuration.rs
eguzki: not sure I follow. `use crate::data::PropertyPath;` cannot be removed. It is being used in this module as part of the `SelectorItem`.

tests/multi.rs
alexsnaps: I wouldn't bother too much, with the changes that @adam-cattermole is bringing in, we'll have more confidence in these tests... It's all _very_ brittle right now. This is probably an area we might want to invest properly in. It'd go beyond this repo tho... i.e. in the test-fw itself too probably.

PR #141, opened 2024-11-12
Move selectors to expressions in dev environment

adam-cattermole:
* Update examples to use expressions instead of selectors (accidentally missed from #137)
* Lowercase the header in the example for `request.headers`
* Add `debug` log for the `request.headers` map

### DIFF
--- a/utils/deploy/envoy.yaml
+++ b/utils/deploy/envoy.yaml
@@ -180,8 +180,9 @@ data:
                                           "scope": "rlp-ns-A/rlp-name-A",
                                           "data": [
                                             {
-                                              "selector": {
-                                                "selector": "unknown.path"
+                                              "expression": {
+                                                "key": "unknown.path",
+                                                "value": "unknown.path"
                                               }
                                             }
                                           ]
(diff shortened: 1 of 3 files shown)


utils/deploy/envoy.yaml
```diff
@@ -237,19 +238,21 @@ data:
...
+                                                "value": "request.headers['My-Custom-Header-01']"
                                               }
                                             },
                                             {
-                                              "selector": {
-                                                "selector": "metadata.filter_metadata.envoy\\.filters\\.http\\.header_to_metadata.user_id",
-                                                "key": "user_id"
+                                              "expression": {
+                                                "key": "user_id",
+                                                "value": "string(getHostProperty(['metadata', 'filter_metadata', 'envoy.filters.http.header_to_metadata', 'user_id']))"
```
eguzki, line 255: shouldn't be `auth.somthing`?

eguzki, line 255: I wonder what would be required if the auth step is configured using gateway capabilities like [Security Policy]( https://gateway.envoyproxy.io/v0.6/design/security-policy/) (but not used for ext auth service, but `jwt` for instance.

adam-cattermole: No this one is coming from the [`header_to_metadata`](https://github.com/Kuadrant/wasm-shim/blob/main/utils/deploy/envoy.yaml#L114-L122) filter proving we can get filter state from non-kuadrant filters

eguzki (approved):

PR #144, opened 2024-11-13
Add RL response headers in correct phase

adam-cattermole:
This passes the `response_headers` back up to the filter layer so that we can add the headers only at the last phase

Adding these headers before we send to the upstream service is not allowed, and if you set `rateLimitHeaders: DRAFT_VERSION_03` for limitador the shim will panic without this change.

### DIFF
--- a/src/filter/http_context.rs
+++ b/src/filter/http_context.rs
@@ -142,13 +142,9 @@ impl Context for Filter {
 
         match op_res {
             Ok(operation) => {
-                if GrpcService::process_grpc_response(
-                    operation,
-                    resp_size,
-                    &mut self.response_headers_to_add,
-                )
-                .is_ok()
-                {
+                if let Ok(result) = GrpcService::process_grpc_response(operation, resp_size) {
+                    // add the response headers
+                    self.response_headers_to_add.extend(result.response_headers);
                     // call the next op
                     match self.operation_dispatcher.borrow_mut().next() {
                         Ok(some_op) => {
--- a/src/filter/http_context.rs
+++ b/src/filter/http_context.rs
@@ -142,7 +142,13 @@ impl Context for Filter {
 
         match op_res {
             Ok(operation) => {
-                if GrpcService::process_grpc_response(operation, resp_size).is_ok() {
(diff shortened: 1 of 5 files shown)


adam-cattermole:
Apologies to @didierofrivia here I incorrectly suggested to break this https://github.com/Kuadrant/wasm-shim/pull/95#discussion_r1787664108

src/filter/http_context.rs
```diff
@@ -142,7 +142,13 @@ impl Context for Filter {
 
         match op_res {
             Ok(operation) => {
-                if GrpcService::process_grpc_response(operation, resp_size).is_ok() {
+                if GrpcService::process_grpc_response(
+                    operation,
+                    resp_size,
+                    &mut self.response_headers_to_add,
```
alexsnaps, line 148: This is a code smell... this seems to invert the indirection between the state machine possibly progressing and the result of such a progress... Not a huge deal, I don't think this is even a first. "Just" mentioning it... 

The way I think about it: something happens, this has a "result" (i.e. some output) and that output is dealt with. Which in this particular case would be adding the headers to this `Vec`, but the action knows nothing about that `Vec`, instead it merely returns the entries to add to a caller that then knows what to do with them.

alexsnaps (approved):

src/filter/http_context.rs
adam-cattermole: Right, so you think it'd be preferable to have some kind of `GrpcResult` containing the code and the response headers to let the filter deal with that

alexsnaps: Never mind for now... we probably would want to go to the drawing board at some point and design this a little more thoroughly

adam-cattermole: I missed this.. I realise we would likely change a lot of things here, but how about this alternative in latest commit?

didierofrivia (approved):
> Apologies to @didierofrivia here I incorrectly suggested to break this [#95 (comment)](https://github.com/Kuadrant/wasm-shim/pull/95#discussion_r1787664108)

No worries, at the time it made sense to me too! 

👍🏼

PR #147, opened 2024-11-14
Replace all unwraps

adam-cattermole:
* Enable clippy lint `unwrap_used` - this is in a separate commit so we can drop it before we merge
* Most cases was replacing unwraps with expect

### DIFF
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -106,10 +106,7 @@ impl TryFrom<PluginConfiguration> for FilterConfig {
                     .expect("Predicates must not be compiled yet!");
 
                 for datum in &action.data {
-                    let result = datum.item.compile();
-                    if result.is_err() {
-                        return Err(result.err().unwrap());
-                    }
+                    datum.item.compile()?;
                 }
             }
 
@@ -204,7 +201,10 @@ impl<'de> Visitor<'de> for TimeoutVisitor {
         E: Error,
     {
         match duration(Arc::new(string)) {
-            Ok(Value::Duration(duration)) => Ok(Timeout(duration.to_std().unwrap())),
+            Ok(Value::Duration(duration)) => duration
+                .to_std()
+                .map(Timeout)
+                .map_err(|e| E::custom(e.to_string())),
             Err(e) => Err(E::custom(e)),
             _ => Err(E::custom("Unsupported Duration Value")),
         }
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -277,7 +277,7 @@ mod test {
         }
         assert!(res.is_ok());
 
-        let filter_config = res.unwrap();
+        let filter_config = res.expect("result is ok");
         assert_eq!(filter_config.action_sets.len(), 1);
 
         let services = &filter_config.services;
@@ -364,7 +364,7 @@ mod test {
         }
         assert!(res.is_ok());
 
-        let filter_config = res.unwrap();
+        let filter_config = res.expect("result is ok");
         assert_eq!(filter_config.action_sets.len(), 0);
     }
 
@@ -410,12 +410,15 @@ mod test {
         }
         assert!(res.is_ok());
 
-        let filter_config = res.unwrap();
+        let filter_config = res.expect("result is ok");
         assert_eq!(filter_config.action_sets.len(), 1);
 
         let services = &filter_config.services;
         assert_eq!(
-            services.get("limitador").unwrap().timeout,
+            services
+                .get("limitador")
+                .expect("limitador service to be set")
+                .timeout,
             Timeout(Duration::from_millis(20))
         );
 
@@ -510,7 +513,7 @@ mod test {
         }
         assert!(res.is_ok());
 
-        let result = FilterConfig::try_from(res.unwrap());
+        let result = FilterConfig::try_from(res.expect("result is ok"));
         let filter_config = result.expect("That didn't work");
         let rlp_option = filter_config
             .index
--- a/src/data/attribute.rs
+++ b/src/data/attribute.rs
@@ -166,7 +166,10 @@ fn process_metadata(s: &Struct, prefix: String) -> Vec<(String, String)> {
             let nested_struct = value.get_struct_value();
             result.extend(process_metadata(nested_struct, current_prefix));
         } else if let Some(v) = json {
-            result.push((current_prefix, serde_json::to_string(&v).unwrap()));
+            match serde_json::to_string(&v) {
+                Ok(ser) => result.push((current_prefix, ser)),
+                Err(e) => error!("failed to serialize json Value: {e:?}"),
+            }
         }
     }
     result
--- a/src/data/cel.rs
+++ b/src/data/cel.rs
@@ -7,8 +7,10 @@ use cel_parser::{parse, Expression as CelExpression, Member, ParseError};
 use chrono::{DateTime, FixedOffset};
 #[cfg(feature = "debug-host-behaviour")]
 use log::debug;
+use log::warn;
 use proxy_wasm::types::{Bytes, Status};
 use serde_json::Value as JsonValue;
+use std::borrow::Cow;
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::fmt::{Debug, Formatter};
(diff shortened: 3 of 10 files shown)


src/configuration.rs
```diff
@@ -204,7 +201,11 @@ impl<'de> Visitor<'de> for TimeoutVisitor {
         E: Error,
     {
         match duration(Arc::new(string)) {
-            Ok(Value::Duration(duration)) => Ok(Timeout(duration.to_std().unwrap())),
+            Ok(Value::Duration(duration)) => Ok(Timeout(
+                duration
+                    .to_std()
+                    .expect("duration should not be less than zero"),
```
alexsnaps, line 207: Should this return a `Result.Err` instead?

src/data/attribute.rs
```diff
@@ -166,7 +166,10 @@ fn process_metadata(s: &Struct, prefix: String) -> Vec<(String, String)> {
             let nested_struct = value.get_struct_value();
             result.extend(process_metadata(nested_struct, current_prefix));
         } else if let Some(v) = json {
-            result.push((current_prefix, serde_json::to_string(&v).unwrap()));
+            result.push((
+                current_prefix,
+                serde_json::to_string(&v).expect("failed to serialize json Value"),
```
alexsnaps, line 171: Would this contain any information about `v`? If not, maybe `.inspect` or otherwise use the error to log?
I'd maybe even go as far as dropping that mapping on the floor, log at error! Or explicitly add a `null` value or... rather than having the whole process crash?

src/data/cel.rs
```diff
@@ -102,8 +102,16 @@ fn decode_query_string(This(s): This<Arc<String>>, Arguments(args): Arguments) -
     for part in s.split('&') {
         let mut kv = part.split('=');
         if let (Some(key), Some(value)) = (kv.next(), kv.next().or(Some(""))) {
-            let new_v: Value = decode(value).unwrap().into_owned().into();
-            match map.entry(decode(key).unwrap().into_owned().into()) {
+            let new_v: Value = decode(value)
+                .expect("failed to decode query string value")
```
alexsnaps, line 106: Maybe same reasoning as the json encoding applies here, wdyt?

src/data/cel.rs
```diff
@@ -296,11 +309,11 @@ fn json_to_cel(json: &str) -> Value {
             JsonValue::Bool(b) => b.into(),
             JsonValue::Number(n) => {
                 if n.is_u64() {
-                    n.as_u64().unwrap().into()
+                    n.as_u64().expect("number must be u64").into()
```
alexsnaps, line 312: You could maybe prefix that string with `Unreachable: `, as the contract is that if `.is_<T>`, `.as_<T>` succeeds

src/service.rs
```diff
@@ -57,7 +57,8 @@ impl GrpcService {
     ) -> Result<GrpcResult, StatusCode> {
         let failure_mode = operation.get_failure_mode();
         if let Some(res_body_bytes) =
-            hostcalls::get_buffer(BufferType::GrpcReceiveBuffer, 0, resp_size).unwrap()
+            hostcalls::get_buffer(BufferType::GrpcReceiveBuffer, 0, resp_size)
+                .expect("failed to get_buffer on GrpcReceiveBuffer")
```
alexsnaps, line 61: Should this leverage `FailureMode`?

src/service.rs
```diff
@@ -140,7 +143,8 @@ impl GrpcServiceHandler {
         message: GrpcMessageRequest,
         timeout: Duration,
     ) -> Result<u32, Status> {
-        let msg = Message::write_to_bytes(&message).unwrap();
+        let msg = Message::write_to_bytes(&message)
+            .expect("failed to write the GrpcMessageRequest to bytes");
```
alexsnaps, line 147: `FailureMode` again?

src/data/cel.rs
alexsnaps: Tho, here, I think I'd just use the undecoded string instead when that happens

src/data/cel.rs
adam-cattermole: Yeah that's clearer, that was kinda what I was trying to convey in my message.. that the number cannot be anything other than the type

src/service.rs
adam-cattermole: Yep good point, this and below should be handled

src/data/attribute.rs
chirino: In what cases would the serialization fail?

alexsnaps: Right, given the above, _it can only be_ `String`, `Bool`, `null` or some Number... so I guess it shouldn't ever fail indeed... for as long as that invariant remains true tho...

src/data/cel.rs
adam-cattermole: Done

src/lib.rs
```diff
@@ -1,6 +1,7 @@
+#![deny(clippy::all, clippy::unwrap_used)]
 mod configuration;
 mod data;
-#[allow(renamed_and_removed_lints)]
+#[allow(renamed_and_removed_lints, clippy::unwrap_used)]
```
alexsnaps, line 4: So, wdyt? Should we remove it? Or shall we keep it?
It's more a question for the team tbh @didierofrivia @eguzki 
If we leave it, we need to all now _always_ really consider the error handling, which is probably a good thing, but leads to more work. If we remove it, I think we should redo this exercise regularly... 
Now leaving it doesn't mean use `.expect("I need this!")` instead of `.unwrap()` always obviously! :P

src/service/auth.rs
```diff
@@ -64,7 +64,7 @@ impl AuthService {
         let mut request = AttributeContext_Request::default();
         let mut http = AttributeContext_HttpRequest::default();
         let headers: HashMap<String, String> = hostcalls::get_map(MapType::HttpRequestHeaders)
-            .unwrap()
+            .expect("failed to retrieve HttpRequestHeaders from host")
```
alexsnaps, line 67: Let's leave this for now (and the other 2 calls below, as well as the one in `rate_limit.rs`), but I think this should eventually become some kind of `FailureMode` dependent behavior...

alexsnaps (approved):
But let's wait on what others in the team think wrt the lint

src/service/auth.rs
adam-cattermole: 100% agreed again here

src/lib.rs
didierofrivia: I think it makes sense to remove it since we want fast iteration... we could start talking about code quality and practices once the storm passes. "Make it ugly, make it work" first, but we need to compromise to make it pretty and robust afterwards.

src/service/auth.rs
andrewdavidmackenzie: The expect adds some info, but still causes a panic...which is what we're looking to avoid, no?

andrewdavidmackenzie: maybe unwrap_or() and provide some sane default (an empty map?)?

src/lib.rs
adam-cattermole: I've removed the lint for now, we can re-enable this whenever we like, but should revisit this in any case

src/service/auth.rs
adam-cattermole: The primary problem is this should perform an action based on the `FailureMode` defined for the service - if this service is set to `Deny` then we don't want to populate the request with a default value and let it through. This is a much larger refactor though than can be solved in this PR

didierofrivia (approved):
👍🏼

alexsnaps:
>if we were going to support more the library model) stabilized the API, as I expected a lot of return values to have to change to Result<>

This is definitively true for Limitador... here, things a slightly different tho. But _yes_ this _is more work_, yet useful work... _if we don't just replace `.unwrap()` with `.expect()` out of ... wanting to do that work, which is think about the issue hard and what it means to the user. Where in this case the user is a moving target: the kuadrant user, the downstream client, which can be an evil actor (e.g. badly encoded query strings...). Anyways, we _need to all_ agree on this lint being in. Otherwise, I still believe we are better off doing a pass such as this one e.g. every release?

src/service/auth.rs
alexsnaps: >provide some sane default (an empty map?)?

there is no "sane default" in this case, as it will just propagate the error to the next inline... i.e. in this case probably the CEL expression is going to fail. This error here tho, would mean something went wrong with envoy. 

As @adam-cattermole rightly pointed out, this would need to fallback to some user configured behavior (analogous to e.g. Limitador being unreachable). So we need to get rid of this one when we figured out that part of the issue (see [this "issue"](https://github.com/Kuadrant/kuadrant-operator/issues/1031) for more context

Issue #145, opened 2024-11-13
Fold subsequent calls to Limitador into a single one


alexsnaps:

alexsnaps:
@eguzki you said iirc you started on this, but couldn't find anything, is this still on your plate?

eguzki:
Yes! working on this right now.

eguzki:
v1.0 or v1.1? @alexsnaps

alexsnaps:
v1.0... made it a should tho. Don't think it should be a blocker

PR #148, opened 2024-11-18
Fold subsequent calls to limitador into a single one

eguzki:
### What

Addresses #145

Divided the work in two commits:
* Easy changes, as a preparation work for the actual feature. Basically create new runtime structure of config and actions. 
* Fold subsequent calls to limitador into a single one

I moved things around a bit, but let me know if you want to rearrange this differently 🙏


alexsnaps:
>I moved things around a bit, but let me know if you want to rearrange this differently 🙏

I don't know that I would want it differently, but it's certainly surprising naming imho. Now... As a general note, I'd really appreciate if you could keep "moving things around" as a dedicated commit in future PRs, as it make reviewing the actual changes easier and makes for a nicer history too (e.g. `git bisect`, et al...)

Can you explain how you think the module `internal` adds value tho? I'm a little confused by its name (tho it could be me, and how I makes me think of [`Pimpl`](https://en.cppreference.com/w/cpp/language/pimpl) and "`inner`" in Rust, but unsure why). Anyways, I get a little confused about what everyone is trying to achieve with modules in this code base.

eguzki:
Happy to remove the `internal` module and add all the "internal" modules to the root one.  I was trying to make a clear distinction between  `PluginConfiguration` and `FilterConfig`.

eguzki:
`RuntimeConfig` will be

alexsnaps:
> `RuntimeConfig` will be

I'm _not_ saying that's the right approach. _I think_ that _my confusion_ between these types is their naming. I don't know that I see the whole picture. And most importantly I think we should keep these changes separated, at least in separate commits. There is no _ask to change anything_ on my part here, just to be clear. I'm trying to understand the goals and then look at the means used.

eguzki:
@alexsnaps let me know if you want to rearrange this differently 🙏

tests/multi.rs
```diff
@@ -198,15 +198,15 @@ fn it_performs_authenticated_rate_limiting() {
...
+                26, 14, 10, 7, 58, 109, 101, 116, 104, 111, 100, 18, 3, 71, 69, 84, 26, 30, 10, 10,
+                58, 97, 117, 116, 104, 111, 114, 105, 116, 121, 18, 16, 97, 98, 105, 95, 116, 101,
+                115, 116, 95, 104, 97, 114, 110, 101, 115, 115, 26, 38, 10, 5, 58, 112, 97, 116,
+                104, 18, 29, 47, 100, 101, 102, 97, 117, 108, 116, 47, 114, 101, 113, 117, 101,
+                115, 116, 47, 104, 101, 97, 100, 101, 114, 115, 47, 112, 97, 116, 104, 34, 10, 47,
+                97, 100, 109, 105, 110, 47, 116, 111, 121, 42, 17, 99, 97, 114, 115, 46, 116, 111,
+                121, 115, 116, 111, 114, 101, 46, 99, 111, 109, 50, 4, 104, 116, 116, 112, 82, 8,
+                72, 84, 84, 80, 47, 49, 46, 49, 82, 20, 10, 4, 104, 111, 115, 116, 18, 12, 97, 117,
+                116, 104, 99, 111, 110, 102, 105, 103, 45, 65, 90, 0,
             ]),
```
alexsnaps, line 210: Did we make sure these changes are valid? Don't remember now, but didn't @adam-cattermole look into this? I was hoping we'd get to something that makes it clear what these bytes are, but maybe that didn't happen... 🤔

src/service.rs
```diff
@@ -18,7 +18,7 @@ use std::cell::OnceCell;
 use std::rc::Rc;
 use std::time::Duration;
 
-#[derive(Default, Debug)]
+#[derive(Default, Clone, Debug)]
```
alexsnaps, line 21: Why is this cloned now? I see we store `Rc<GrpcService>` in out "actions" 🤔

src/ratelimit_action.rs
```diff
@@ -0,0 +1,332 @@
...
+    grpc_service: Rc<GrpcService>,
+    scope: String,
+    service_name: String,
+    conditional_data_sets: Vec<ConditionalData>,
+}
+
+impl RateLimitAction {
+    pub fn new(action: &Action, service: &Service) -> Result<Self, String> {
+        Ok(Self {
+            grpc_service: Rc::new(GrpcService::new(Rc::new(service.clone()))),
```
alexsnaps, line 138: So that's an `Rc` of ... one _always_?

src/ratelimit_action.rs
```diff
@@ -0,0 +1,332 @@
...
+        true
+    }
+
+    pub fn get_failure_mode(&self) -> FailureMode {
+        self.grpc_service.get_failure_mode()
+    }
+
+    pub fn merge(&mut self, other: &RateLimitAction) {
+        self.conditional_data_sets
+            .extend_from_slice(&other.conditional_data_sets);
```
alexsnaps, line 182: that's why `GrpcService` needs to be `Clone`... here, on merge, it gets cloned and the `Rc` gets incremented. But... Yeah, well... if this works lets keep this for now.

alexsnaps (approved):
Approving this so we make progress. But we really need to schedule some time with @adam-cattermole & @didierofrivia soon enough to start putting some serious thoughts in what it is we want the design of this to become.

tests/multi.rs
adam-cattermole: The ordering of the serialized headers changed again... it was `:authority` -> `:method` -> `:path` and became `:authority` -> `:path` -> `:method`. I don't know why this keeps changing when protobuf serialized but is consistent enough to pass tests for each PR?

src/runtime_action.rs
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    pub fn merge(&mut self, other: &RuntimeAction) {
+        // only makes sense for rate limiting actions
+        if let Self::RateLimit(self_rl_action) = self {
+            if let Self::RateLimit(other_rl_action) = other {
+                self_rl_action.merge(other_rl_action)
+            }
+        }
+    }
```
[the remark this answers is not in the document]
eguzki: Well, not really. `RateLimitAction` can build descriptor entries and implements `merge` method. Also the `conditions_apply` is different because the predicates are now *guards* to allow descriptor entries to be included in the descriptor or not.

tests/multi.rs
eguzki: this is something I still do not understand. It just works, it seems to be consistent, but when something apparently *unrelated* is changed, then the request body changes :shrug:

src/ratelimit_action.rs
eguzki: I will take a look

src/service.rs
eguzki: I do not think it is needed. It *was* for something I tried. Not anymore.

tests/multi.rs
alexsnaps: Not saying this is it but, I know some libs from Google around proto will purposefully shuffle things around in there "serialized" form _when nothing about that order should be asserted for_. One example is actually protobuf json that does change the resulting json literal string around...

eguzki:
@alexsnaps ready for review

Added easy changes in the first commit to make changes easy on the second commit :heart:

eguzki:
@adam-cattermole @didierofrivia you feedback more than welcome as well. Do not hesitate, be sharp. This is not rushing for v1. We want to share mindset of "common goals". I am happy to address any comment

src/ratelimit_action.rs
adam-cattermole: I think similar to my previous comment, we also used to create all the `GrpcService` up front and share them as we know these won't change, now we're still creating them up front on configure time but one for every action?

src/operation_dispatcher.rs
```diff
@@ -171,25 +168,18 @@ impl OperationDispatcher {
...
-                    Default::default(),
-                ));
-            }
+            operations.push(Rc::new(Operation::new(
+                Rc::clone(action),
+                GrpcServiceHandler::new(
+                    Rc::clone(&action.grpc_service()),
+                    Rc::clone(&self.header_resolver),
+                ),
+            )));
```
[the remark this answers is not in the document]
eguzki: > the original implementation passed references to GrpcServiceHandlers and only created as many as we had services, instead of one per operation as we have now

Correct. The reason being that the runtime action now has the grpc_service embedded, as part of it. It could even be the GrpcServiceHandler itself, why not? It happens that creating GrpcServiceHandler is cheap as it is composed of a couple of `Rc<T>` fields. The benefit is to remove the `service_handlers: HashMap<String, Rc<GrpcServiceHandler>>` from the `OperationDispatcher` and remove the need to check at request time if the service exist or not. Each action, which are built at config time, has the assigned service and grpc_service instance ready to be used. For example, if there are 10 services defined in the config, but actions only use 2 of them why keep them  all in `OperationDispatcher` ? 

That is the reasoning. Let me know if that makes sense or not.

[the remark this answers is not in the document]
alexsnaps: Yeah, something is still fishy with these `Rc`s tbh... if anthing tho, this `Rc::clone` is unnecessary, the callee is already doing that, and returning an `Rc` for the caller to own. That just literally just increments the counter by one, to drop the returned reference back on the floor and as such decrementing the counter. That being said the compiler _may_ opt this out... yet, it leave the unnecessary noise for the reader to parse out.

alexsnaps:
General note on the refactoring: the duplication between `RateLimitAction` & `AuthAction`, namely both fields `grpc_service` & `scope` "only" to then be dispatched from `RuntimeAction`, makes me wonder if that's not where these two fields should live, rather than in the specialized actions. They seem to serve no other purpose there than to be exposed to the layer up, which is `RuntimeAction`, right?

```rust
pub struct RateLimitAction {
    grpc_service: Rc<GrpcService>,
    scope: String,
    service_name: String,
    conditional_data_sets: Vec<ConditionalData>,
}
```

```rust 
pub struct AuthAction {
    grpc_service: Rc<GrpcService>,
    scope: String,
    predicates: Vec<Predicate>,
}
```

src/operation_dispatcher.rs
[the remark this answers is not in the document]
eguzki: good catch _alexclippy_

Fixed the unnecessary RC clone

src/ratelimit_action.rs
adam-cattermole: Sure, I think it's fine for now, but I personally think it's wrong to have as many `GrpcService`s as we have actions, with 2 services defined in the config and N actions (assuming no merges) we'll have N services when we had 2 before.

src/ratelimit_action.rs
```diff
@@ -0,0 +1,320 @@
...
+
+    pub fn scope(&self) -> &str {
+        self.scope.as_str()
+    }
+
+    pub fn conditions_apply(&self) -> bool {
+        // For RateLimitAction conditions always apply.
+        // It is when building the descriptor that it may be empty because predicates do not
+        // evaluate to true.
+        true
```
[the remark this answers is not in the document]
eguzki: the `ratelimit_action` instance holds data like this

```yaml
- predicates:
  - 1 == 1
  - 2 == 2
  data:
  - expression:
      key: a
      value: "1"
  - expression:
      key: b
      value: 1
- predicates:
  - 3 == 3
  - 4 == 4
  data:
  - expression:
      key: c
      value: "1"
  - expression:
      key: d
      value: 1
```

So, I had two options:
A) `conditions_apply` method would evaluate _all_ the predicates and return the evaluation result _AND_'ed. The on `build_descriptors` evaluate them **again** and generate descriptor entries.
B)  `conditions_apply` would evaluate always to `true` and  `build_descriptors`  call would evaluate them (only once) and generate descriptor entries. If the descriptor is empty (no entries  generated), that means no need to run grpc call to limitador. 

As you see in the code, I implemented the latter approach. Two reasons: 1) predicates are evaluated _once_ and 2) even if predicates evaluate to `true`, the generation of one entry, theoretically, may not generate one entry. Think about trying to read one headers value when the header does not exists. It is true that the way it is currently implemented, `DescriptorEntryBuilder`'s  `evaluate` method always returns a `RateLimitDescriptor_Entry`. Which I am not entirely sure about that. The current implementation would return `''null''` string when the requested header value does not exist. I believe this should not be the case and it should return `Option<RateLimitDescriptor_Entry>` instead, but this is topic for another discussion. 

I am happy to implement the first option or even an alternative option.

eguzki:
> General note on the refactoring: the duplication between `RateLimitAction` & `AuthAction`, namely both fields `grpc_service` & `scope` "only" to then be dispatched from `RuntimeAction`, makes me wonder if that's not where these two fields should live, rather than in the specialized actions. They seem to serve no other purpose there than to be exposed to the layer up, which is `RuntimeAction`, right?
> 
> ```rust
> pub struct RateLimitAction {
>     grpc_service: Rc<GrpcService>,
>     scope: String,
>     service_name: String,
>     conditional_data_sets: Vec<ConditionalData>,
> }
> ```
> 
> ```rust
> pub struct AuthAction {
>     grpc_service: Rc<GrpcService>,
>     scope: String,
>     predicates: Vec<Predicate>,
> }
> ```

I think you are right about the `grpc_service`. It serves no other purpose there than to be exposed to the layer up. How could be part of the `RuntimeAction` ? Maybe instead of `enum` having `RuntimeAction<T>` ?

Regaring `scope`, this field is used to be exposed to the layer up as well as to build the `GrpcMessageRequest` for which the specialization is needed. [Here](https://github.com/Kuadrant/wasm-shim/blob/main/src/service/grpc_message.rs#L125-L145)

alexsnaps:
>Regaring scope, this field is used to be exposed to the layer up as well as to build the GrpcMessageRequest for which the specialization is needed. [Here](https://github.com/Kuadrant/wasm-shim/blob/main/src/service/grpc_message.rs#L125-L145)

Right, but that function takes an `RuntimeAction` as an argument... so there no need to navigate further down. That `&RuntimeAction` could be the source of the `scope`, just as `grpc_service`. This is unnecessary coupling (and duplication)

adam-cattermole (approved):
I don't want to hold this up, I think we still have the refactor yet to do, and this improves readability, is tested and works for the intended purpose so I'm happy to approve, nice work

I think it just has some of the remnants from the operation dispatcher and previous implementation in the form of the services/handlers/rcs to these but we can iterate

eguzki:
> uld be the source of the `scope`, just as `grpc_service`. T

You are right. _But_ the build the `GrpcMessageRequest` is specific and scope is needed for that job, so it makes all the sense that RatelimitAction and AuthAction know about `scope`. Moreover, it turns out that ratelimit action uses the `scope` as well to implement the `merge`.

alexsnaps (approved):
I'm with @adam-cattermole 
I think the _design_ is lacking mostly, but indeed still very much informed by the existing... that we know we need to give some love

src/ratelimit_action.rs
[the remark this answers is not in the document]
alexsnaps: It's less about the approach per se... tho, yeah that could also be challenged. It's about the abstraction introduced.

alexsnaps:
>But the build the GrpcMessageRequest is specific and scope is needed for that job, so it makes all the sense that RatelimitAction and AuthAction know about scope.

I very much disagree with that statement, as right now it's a split responsibility: `GrpcMessageRequest::new` take a `&RuntimeAction` and then splits... So the coupling is (from a "public" API to `RuntimeAction`) then internally to both variants... Where that `enum` is currently just dynamic dispatching to an "interface" that's ill defined (all "API" from `*Action`s _but_ `Predicate` vs `Conditional` ones...)

eguzki:
For me, instead of having `GrpcMessageRequest::new(action: &RuntimeAction) -> Option<Self>`, it should be  RuntimeAction doing some job which implementation depends on the variant. And that job may (or may not) need to create a GrpcMessageRequest, for which you need the scope.

Anyway... for future refactors.

alexsnaps:
Right... I'm thinking we're under utilizing `GrpcService`, I think there's an inversion here possibly. That is the differentiator again... Too much logic (the dynamic dispatch) in `RuntimeAction` 🤔

src/operation_dispatcher.rs
```diff
@@ -347,27 +331,48 @@ mod tests {
         }
     }
 
-    fn build_operation(
-        grpc_call_fn_stub: GrpcCallFn,
-        extension_type: ServiceType,
-    ) -> Rc<Operation> {
+    fn build_auth_grpc_action() -> RuntimeAction {
```
didierofrivia, line 334: Just leaving a note here, for a future refactor... the `operationDispatcher` was meant to be agnostic of the service/extension specifics. Anything specific for auth, ratelimit, what ever else that comes in the future, should be built somewhere else.

PR #157, opened 2024-12-04
Run next action on failure when failuremode is allow

### LINKED ISSUE #149: FailureMode propagation and handling in services
There is a flaw in the logic to the handling of errors for each of the services - we should always propagate the error to the caller for the next action in the set.

Example; if there's two services defined, the first action being `FailureMode::Allow` followed by `FailureMode::Deny` and for any reason we have an issue in the [`process_grpc_response`](https://github.com/Kuadrant/wasm-shim/blob/main/src/service.rs#L55) function for the first service we call [`handle_error_on_grpc_response`](https://github.com/Kuadrant/wasm-shim/blob/main/src/service.rs#L85-L95) which will [`resume_http_request`](https://github.com/Kuadrant/wasm-shim/blob/main/src/service.rs#L92) ignoring the subsequent actions effectively allowing the request through.

This [error block](https://github.com/Kuadrant/wasm-shim/blob/main/src/service.rs#L70-L76) for example will also be hit if we have a timeout on our grpc call to the initial service (service code 14 with an empty message body).

eguzki:
### What 

Fixes #149


Cargo.lock
```diff
@@ -1,6 +1,6 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 3
+version = 4
```
adam-cattermole, line 3: I think this is due to a newer cargo version as my local one doesn't support this, we are pinned at 1.75.0 for local development

tests/failuremode.rs
```diff
@@ -0,0 +1,168 @@
...
+        .returning(Some(&[]))
+        .expect_grpc_call(
+            Some("limitador-cluster"),
+            Some("envoy.service.ratelimit.v3.RateLimitService"),
+            Some("ShouldRateLimit"),
+            Some(&[0, 0, 0, 0]),
+            Some(&[
+                10, 1, 97, 18, 28, 10, 26, 10, 21, 108, 105, 109, 105, 116, 95, 116, 111, 95, 98,
+                101, 95, 97, 99, 116, 105, 118, 97, 116, 101, 100, 18, 1, 49, 24, 1,
+            ]),
```
adam-cattermole, line 144: Do we want to use `Some` vs `None` here and are the bytes here correct?

tests/failuremode.rs
```diff
@@ -0,0 +1,168 @@
...
+        .unwrap();
+
+    let http_context = 2;
+    module
+        .call_proxy_on_context_create(http_context, root_context)
+        .expect_log(Some(LogLevel::Debug), Some("#2 create_http_context"))
+        .execute_and_expect(ReturnType::None)
+        .unwrap();
+
+    let first_call_token_id = 42;
```
adam-cattermole, line 94: Should we just leave this out and have it hard-coded the same as the second token? or use a diff token than 42 for both?

tests/failuremode.rs
eguzki: `None`!

eguzki:
handed over to #158

PR #161, opened 2025-01-16
Refactor entire filter flow

adam-cattermole:
# Changes

In this PR I've refactored the existing approach which had become somewhat unwieldy. I have tried to clean up the commits here, but it's worth noting this was an iterative process, and so some of the changes in the initial commits are changed/reverted in subsequent ones.

The main principle is as follows - I have aimed to reduce the leaking of the wasm/envoy hostcalls across every layer and instead delegate this to the filter. The filter has knowledge of the different 'phases' (`on_http_request.. on_grpc_call_response..`), and simply just performs `Operations` based on it's current state and what it is being asked to do. The configuration itself remains unchanged, and the concerns of building and consuming of the specialised message types is delegated to the owner (`auth_action/rl_action`). The filter has no knowledge of the individual `actions` but instead uses the `RuntimeActionSet` to initiate the flow, and from then on the `Operations` to determine what to do next.

Each of the [operations](https://github.com/Kuadrant/wasm-shim/blob/proposal-refactor/src/filter/operations.rs#L7-L14) tell the filter it must do something (that the layers below should not and do not know about):

* `SendGrpcRequest`: This informs the filter it needs to make a GRPC call and contains all of the the information it requires to make the request - the filter does not need to know which action/service this relates to
* `AwaitGrpcResponse`: This tells the filter that there is nothing left to do within the current method and it must pause to await a GRPC response
* `AddHeaders`: This tells the filter it must add headers to the response in the required phase
* `Die`: This informs the filter that it is finished due to an error and must send a response before it finishes - this may not be an "error" due to a failure, but is also used for the case where we must return an error code to the caller (i.e. 401 Forbidden, 429 Too Many Requests..)
* `Done`: This informs the filter that it has successfully finished operating and the request may be resumed

The specialised actions currently handle both the building of messages to the transfer type, and conversion to the message they expect. I have also made some decisions on how these actions return information to the layers above, based on the different operations the filter may have to perform.
### Outstanding

There are several outstanding things to address (in follow up work):

* Currently some parts of the logic are based around whether there is a message to send or not, this could be reworked to determine whether we have *anything* to do, for example our action just requires us to add headers and not send a message
* Whilst the hostcalls have been removed in lots of places throughout the code, there are still uses through the `set/get_property` calls, this would ideally be replaced with some kind of resolver layer with memoization
* To keep an abstraction between the `KuadrantFilter`, `Operations`, and `RuntimeActions` there was a need to retain the index of the current active `RuntimeAction` within some `Operations` (to progress the flow). I would like to investigate eliminating this, potentially by building the chain of `Operations` up front (where each contains self + next). Another alternative might be implementing the iterator trait so we can just call `.next()` to go through each


alexsnaps:
I feel I'm not the right person to review this, as I've been kept in the loop all along. So needless to say its a 👍 from me.
Is it perfect? Probably not... Are there bugs? Possibly yes... But it doesn't matter, it's all _in my opinion_ a whole cleaner than it was! And we can keep on iterating as we add new features & capabilities to this component. 

What might be useful is for people to use to PR to record and track (in issues?) what the agreed upon next steps would be? If this is an acceptable starting points for others too obviously.

eguzki (approved):
LGTM

Let's follow this approach and iterate for further improvement

src/filter/kuadrant_filter.rs
```diff
@@ -0,0 +1,187 @@
...
+        );
+        let receiver = mem::take(&mut self.grpc_message_receiver_operation)
+            .expect("We need an operation pending a gRPC response");
+
+        let mut ops = Vec::new();
+
+        if status_code != Status::Ok as u32 {
+            ops.push(receiver.fail());
+        } else if let Some(response_body) = self.get_grpc_call_response_body(0, resp_size) {
+            ops.extend(receiver.digest_grpc_response(&response_body));
```
didierofrivia, line 36: So we need a vec because when digesting the grpc response we might need to add an extra operation (AddHeaders) before sending the next gprc request, right?

adam-cattermole: Yep exactly, right now this can return two ops and we process them in sequence, based on how this all currently works. However this was more so to open the seam to allow us to change how the stuff beneath it works - for example what if we have 5 actions that just want to add headers before continuing? we can just build a vec and let the filter process them all until it needs to 'wait'

didierofrivia (approved):
I liked the path the refactored took, the separation of concerns is more tangible and now it's easier to isolate for unit testing. Still probably an effort to encapsulate hostcalls, maybe we could even write unit tests for the kuadrant filter that way. Great job! thanks! 🎖️

PR #162, opened 2025-01-31
Enable merge queue

adam-cattermole:
Being tracked as part of https://github.com/Kuadrant/kuadrant/issues/7

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -3,11 +3,11 @@ name: Tests
 
 on:
   push:
-    branches:
-      - 'main'
+    branches: ['main']
   pull_request:
-    branches:
-      - '*'
+    branches: ['*']
+  merge_group:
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yaml
```diff
@@ -3,11 +3,12 @@ name: Tests
 
 on:
   push:
-    branches:
-      - 'main'
+    branches: ['main']
```
eguzki, line 6: just out of curiosity, why this change?

laurafitzgerald:
Looks good to me. Happy to approve after questions are finished.

.github/workflows/tests.yaml
adam-cattermole: In my eyes it's either:
```
  pull_request:
    branches:
      - '*'
  merge_group:
    types:
      - checks_requested
```

or the option I opted for so that all the lists are the same format.

PR #165, opened 2025-02-05
Add HeaderKind to enable setting request headers

### LINKED ISSUE #164: Success response headers are not added to the response
**Describe the bug**
Since kuadrant-operator-catalog nightly builds from `28-29.01.25` our tests that should return custom success response headers started failing. They don't add these headers into the response anymore. Here is the [most simple test](https://github.com/Kuadrant/testsuite/blob/4b95ca14fe0ee697f8f0e615adbb0be21a49071d/testsuite/tests/singlecluster/authorino/response/test_simple_response.py#L19) for responses we got that is failing with `AssertionError: Header from response (Header) is missing; assert None is not None`

We were experiencing the same failures a few months ago, which ended up being a bug inside the wasm-shim, and was fixed by [this](https://github.com/Kuadrant/wasm-shim/pull/143) PR.

**Help us Reproduce it**

AuthPolicy:
```yaml
spec:
  rules:
    authentication:
      default:
        credentials:
          authorizationHeader:
            prefix: Bearer
        jwt:
          issuerUrl: 'http://1.2.3.4:8080/realms/realm-averevki--eyt7'
          ttl: 0
        metrics: false
        priority: 0
    response:
      success:
        headers:
          header:
            json:
              properties:
                anything:
                  value: one
            metrics: false
            priority: 0
  targetRef:
    group: gateway.networking.k8s.io
    kind: HTTPRoute
    name: route-averevki--obnu
```
<details>
<summary>Gateway</summary>

```yaml
spec:
  gatewayClassName: istio
  listeners:
    - allowedRoutes:
        namespaces:
          from: Same
      hostname: '*.apps.kua.redhat.com'
      name: api
      port: 80
      protocol: HTTP
```
</details>

<details>
<summary>HTTPRoute</summary>

```yaml
spec:
  hostnames:
    - hostname-averevki--nbep-kuadrant.apps.kua.redhat.com
  parentRefs:
    - group: gateway.networking.k8s.io
      kind: Gateway
      name: gw-averevki--ljiu
  rules:
    - backendRefs:
        - group: ''
          kind: Service
          name: httpbin-averevki--ihex
          namespace: kuadrant
          port: 8080
          weight: 1
      matches:
        - path:
            type: PathPrefix
            value: /
```
</details>

**Expected behavior**

Return headers I specify inside AuthPolicy with a successful response

**Environment (please complete the following information):**

kuadrant-operator image: `nightly-04-02-2025`
authorino-operator image: `e1dc399c6bcb2fb6d28b2f3e94ce49bb5b63e06c`
authorino image: `b2bde91bfc64b1d140f740069ba36ce81a3f3349`
wasm-shim image: `751af05ea5e6c440d6254480d669bc5861f457eb`

adam-cattermole:
In the wasm-shim refactor the request vs response headers that was fixed prior (#143) was missed. I've introduced an enum called `HeaderKind` which explicitly differentiates the `Request` vs `Response` types for the specific actions, and have enforced that these can only be added at the correct phase.

I've updated the authconfig to add request headers, so this is verifiable as follows:

```sh
make local-setup
```
```sh
kubectl port-forward --namespace kuadrant-system deployment/envoy 8000:8000
```
```sh
curl -H "Host: test.a.auth.com" -H "Authorization: APIKEY IAMALICE" http://127.0.0.1:8000/get -i
```

See `My-Header` in the **_echoed_** request:
```sh
HTTP/1.1 200 OK
content-type: application/json
server: envoy
date: Wed, 05 Feb 2025 16:10:52 GMT
content-length: 545
x-envoy-upstream-service-time: 2

{
  "method": "GET",
  "path": "/get",
  "query_string": null,
  "body": "",
  "headers": {
    "Host": "test.a.auth.com",
    "User-Agent": "curl/8.7.1",
    "Accept": "*/*",
    "Authorization": "APIKEY IAMALICE",
    "X-Forwarded-For": "10.244.0.10",
    "X-Forwarded-Proto": "http",
    "X-Envoy-Internal": "true",
    "X-Request-Id": "9641a263-bfdb-49b7-b1c2-a7662ec1e74e",
    "My-Header": "{\"my-key\":true}",
    "X-Envoy-Expected-Rq-Timeout-Ms": "15000",
    "Version": "HTTP/1.1"
  },
  "uuid": "fbbaad8b-2d15-42b1-9c3d-c21721d37b59"
}
```

Resolves #164

### DIFF
--- a/src/filter/kuadrant_filter.rs
+++ b/src/filter/kuadrant_filter.rs
@@ -1,9 +1,9 @@
 use crate::action_set_index::ActionSetIndex;
 use crate::filter::operations::{
-    GrpcMessageReceiverOperation, GrpcMessageSenderOperation, HeadersOperation, Operation,
+    GrpcMessageReceiverOperation, GrpcMessageSenderOperation, Operation,
 };
 use crate::runtime_action_set::RuntimeActionSet;
-use crate::service::{GrpcErrResponse, GrpcRequest, HeaderResolver};
+use crate::service::{GrpcErrResponse, GrpcRequest, HeaderResolver, Headers};
 use log::{debug, warn};
 use proxy_wasm::traits::{Context, HttpContext};
 use proxy_wasm::types::{Action, Status};
@@ -16,7 +16,8 @@ pub(crate) struct KuadrantFilter {
     header_resolver: Rc<HeaderResolver>,
 
     grpc_message_receiver_operation: Option<GrpcMessageReceiverOperation>,
-    headers_operations: Vec<HeadersOperation>,
+    response_headers_to_add: Option<Headers>,
+    request_headers_to_add: Option<Headers>,
 }
 
 impl Context for KuadrantFilter {
@@ -51,6 +52,9 @@ impl HttpContext for KuadrantFilter {
         #[cfg(feature = "debug-host-behaviour")]
         crate::data::debug_all_well_known_attributes();
 
+        // default action if we find no action_set where conditions apply
+        let mut action = Action::Continue;
+
         if let Some(action_sets) = self
             .index
             .get_longest_match_action_sets(self.request_authority().as_ref())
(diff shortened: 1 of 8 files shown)


src/filter/kuadrant_filter.rs
```diff
@@ -16,7 +16,8 @@ pub(crate) struct KuadrantFilter {
     header_resolver: Rc<HeaderResolver>,
 
     grpc_message_receiver_operation: Option<GrpcMessageReceiverOperation>,
-    headers_operations: Vec<HeadersOperation>,
+    response_headers_to_add: Option<Headers>,
+    request_headers_to_add: Option<Headers>,
```
adam-cattermole, line 20: These have been renamed and are `Option` that is initialised once and extended until it is taken. Crucially if this is `None` it can no longer be initialised, we have added the headers and we are no longer in the correct phase

src/service.rs
```diff
@@ -131,6 +131,21 @@ impl GrpcRequest {
...
+    Response(Headers),
+    Request(Headers),
+}
+
+impl HeaderKind {
+    pub fn is_empty(&self) -> bool {
+        match self {
+            HeaderKind::Response(headers) => headers.is_empty(),
+            HeaderKind::Request(headers) => headers.is_empty(),
+        }
```
didierofrivia, line 145: You don't need to change this, but if you were wondering how to do it in one liner: 
```suggestion
        match self {
            HeaderKind::Response(headers) | HeaderKind::Request(headers) => headers.is_empty(),
        }

        }
```

adam-cattermole: Oh yeah forgot about that, will add that as it's nicer

didierofrivia (approved):
🍸

PR #167, opened 2025-03-04
Remove unused glob module

adam-cattermole:
Looks like this module survived the refactor and is entirely unused, removing, but if we find we need it we can always find it in git history


didierofrivia (approved):
🔥

PR #169, opened 2025-03-12
Error handling

adam-cattermole:
## Changes

I have tried to improve our error handling to reduce the occurrences of panics, which has resulted in the introduction of several new types
* Existing mapping of errors to `String`s (losing all type information) is gone, and instead we pass errors as their types up the layers
* The panics at lower levels are passed up to the action layer where it decides based on failure mode the respective result
* The filter layer is still informed whether to exit or not by the action beneath it
* ~~I have 'safely' unwrapped a couple of the proxy_wasm hostcalls by calling the underlying functions instead of the trait~~; turns out in these cases it doesn't return an `Err()` but actually `panic!`s at the proxy wasm layers below

### Outstanding:

* I did not choose to 'safely' unwrap `add_request|response_headers`, `resume_http_request`, or `send_http_response` calls as I have been thinking about the expected behaviour here:
    1. If we fail to add headers to the request or response what is the expected outcome? I think we should internal server error regardless of the action failure_mode?
    2. If we fail to `resume_http_request` we could try internal server error but at this stage something is very wrong so perhaps it is okay to panic?
    3. If we fail to `send_http_response` this means we could not send an internal server error... in this case we probably shouldn't silently live where we cannot tell the requester something is wrong
    * For all of the above I am considering a case where this consistently fails, whereas I guess for 2-3 if this is only an error for a single request perhaps we shouldn't panic and let the vm live?

## Verification:

Two of our existing README examples panic, but with the new changes you'll see error logging and an internal server error returned.

Follow the instructions [here](https://github.com/Kuadrant/wasm-shim/?tab=readme-ov-file#running-local-development-environment-kind).

Setup dev env:
```sh
make local-setup
kubectl port-forward --namespace kuadrant-system deployment/envoy 8000:8000
kubecolor logs -f -n kuadrant-system deployments/envoy
```

Make a request:
```sh
curl -H "Host: test.a.rlp.com" http://127.0.0.1:8000/get -i
#HTTP/1.1 500 Internal Server Error
```

Log:
```sh
[2025-03-14 14:31:05.294][45][error][wasm] [source/extensions/common/wasm/context.cc:1201] wasm log kuadrant_wasm kuadrant_wasm vm.sentinel.kuadrant_wasm: Failed to evaluate `Expression { attributes: [Attribute { path: ["unknown", "path"] }], expression: Member(Ident("unknown"), Attribute("path")), extended: false }`: CelError::Resolve { UndeclaredReference("unknown") }
```

Similarly with rlp C, if you did not include the header the request will panic on main:
```sh
curl -H "Host: test.c.rlp.com"  http://127.0.0.1:8000/get -i
#HTTP/1.1 500 Internal Server Error
```

```sh
curl -H "Host: test.c.rlp.com" -H "x-forwarded-for: 50.0.0.1" -H "my-custom-header-01: my-custom-header-value-01" -H "x-dyn-user-id: bob" http://127.0.0.1:8000/get -i
#HTTP/1.1 200 OK
```

### DIFF
--- a/src/auth_action.rs
+++ b/src/auth_action.rs
@@ -197,15 +197,13 @@ mod test {
     #[test]
     fn empty_predicates_do_apply() {
         let auth_action = build_auth_action_with_predicates(Vec::default());
-        assert!(auth_action.conditions_apply().is_ok());
-        assert!(auth_action.conditions_apply().expect("is ok"));
+        assert_eq!(auth_action.conditions_apply(), Ok(true));
     }
 
     #[test]
     fn when_all_predicates_are_truthy_action_apply() {
         let auth_action = build_auth_action_with_predicates(vec!["true".into(), "true".into()]);
-        assert!(auth_action.conditions_apply().is_ok());
-        assert!(auth_action.conditions_apply().expect("is ok"));
+        assert_eq!(auth_action.conditions_apply(), Ok(true));
     }
 
     #[test]
@@ -216,8 +214,7 @@ mod test {
             "true".into(),
             "false".into(),
         ]);
-        assert!(auth_action.conditions_apply().is_ok());
-        assert!(!auth_action.conditions_apply().expect("is ok"));
+        assert_eq!(auth_action.conditions_apply(), Ok(false));
     }
 
     #[test]
--- a/src/auth_action.rs
+++ b/src/auth_action.rs
@@ -1,7 +1,9 @@
 use crate::configuration::{Action, FailureMode, Service};
-use crate::data::{store_metadata, Predicate, PredicateVec};
+use crate::data::{store_metadata, Predicate, PredicateResult, PredicateVec};
 use crate::envoy::{CheckResponse, CheckResponse_oneof_http_response, HeaderValueOption};
+use crate::runtime_action::ResponseResult;
 use crate::service::{GrpcErrResponse, GrpcService, HeaderKind, Headers};
+use cel_parser::ParseError;
 use log::{debug, warn};
 use std::rc::Rc;
 
@@ -13,10 +15,10 @@ pub struct AuthAction {
 }
 
 impl AuthAction {
-    pub fn new(action: &Action, service: &Service) -> Result<Self, String> {
+    pub fn new(action: &Action, service: &Service) -> Result<Self, ParseError> {
         let mut predicates = Vec::default();
         for predicate in &action.predicates {
-            predicates.push(Predicate::new(predicate).map_err(|e| e.to_string())?);
+            predicates.push(Predicate::new(predicate)?);
         }
 
         Ok(AuthAction {
@@ -34,16 +36,15 @@ impl AuthAction {
         self.scope.as_str()
     }
 
-    pub fn conditions_apply(&self) -> bool {
-        //todo(adam-cattermole): do not expect
-        self.predicates.apply().expect("REMOVE")
+    pub fn conditions_apply(&self) -> PredicateResult {
+        self.predicates.apply()
     }
 
     pub fn get_failure_mode(&self) -> FailureMode {
         self.grpc_service.get_failure_mode()
     }
 
-    pub fn resolve_failure_mode(&self) -> Result<HeaderKind, GrpcErrResponse> {
+    pub fn resolve_failure_mode(&self) -> ResponseResult {
         match self.get_failure_mode() {
             FailureMode::Deny => Err(GrpcErrResponse::new_internal_server_error()),
             FailureMode::Allow => {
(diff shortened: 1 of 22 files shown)


src/runtime_action.rs
```diff
@@ -15,11 +18,60 @@ pub enum RuntimeAction {
     RateLimit(RateLimitAction),
 }
 
+pub(super) mod errors {
+    use cel_parser::ParseError;
+    use std::fmt::{Debug, Display, Formatter};
+
+    #[derive(Debug)]
+    pub enum NewActionError {
```
didierofrivia, line 26: I know this error is meant specifically for the creation of actions, but what would you say to rename it to `RuntimeActionError` or simply `ActionError`, the former will tell a better story for the actual mod that is issuing the error and the later would be open for a wider scope.

adam-cattermole: fwiw I _think_ there should not be an overlap between other types of errors because the create errors happen on configuration time whereas other errors relating to runtime actions happen at runtime, so I can't imagine a case where you'd need an enum for a function that can output both

alexsnaps: +1 to renaming it `CreateActionError`, `ActionCreationError` or `ActionIntializationError` or something along those lines. `NewActionError` sounds more like an "order", a verb, than a noun. My $.02

adam-cattermole: Gone with `ActionCreationError`, thanks for the suggestions

src/data/attribute.rs
```diff
@@ -1,137 +1,191 @@
...
 
 impl AttributeValue for i64 {
-    fn parse(raw_attribute: Vec<u8>) -> Result<Self, String> {
-        if raw_attribute.len() != 8 {
-            return Err(format!(
-                "parse: Int value expected to be 8 bytes, but got {}",
-                raw_attribute.len()
-            ));
+    fn parse(raw_attribute: Vec<u8>) -> Result<Self, PropError> {
+        let bytes: Result<[u8; 8], _> = raw_attribute[..8].try_into();
```
eguzki, line 81: what if `raw_attribute` has less than 8 bytes?

adam-cattermole: On the line below we check that the conversion is `ok()` and `== 8`: the logic here has been inverted so that the default base case is to error and in the success case returns the value

alexsnaps: Yeah, no... that's wrong. @eguzki is right, you need to check that it's 8 bytes, _before_ doing the `raw_attribyte[..8]` on line 81, cause [otherwise that'd panic](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=1415c108691a58d41aee7953a9a71f11), as it was doing before. The `Result` is from the `.try_into()`, not the getting the slice into the vector.

alexsnaps: ... or you can just:

`let bytes: Result<[u8; 8], _> = raw_attribute.try_into()`, as that would _only_ work for an 8 byte vector.

adam-cattermole: Ah yeah you're right, I'll address

src/data/cel.rs
```diff
@@ -273,46 +362,39 @@ impl Clone for Attribute {
...
         match &self.cel_type {
             Some(t) => match t {
-                ValueType::String => get_attribute::<String>(&self.path)
-                    .expect("Failed getting to known attribute")
+                ValueType::String => Ok(get_attribute::<String>(&self.path)?
                     .map(|v| Value::String(v.into()))
-                    .unwrap_or(Value::Null),
-                ValueType::Int => get_attribute::<i64>(&self.path)
-                    .expect("Failed getting to known attribute")
+                    .unwrap_or(Value::Null)),
```
alexsnaps, line 370: I think this was wrong all along. if `self.cel_type` is e.g. `String`, we _expect_ a `String`, and in cel this can't be `Null`... now, I wonder if we should add a `self.nullable: bool` and use that to either fallback to `Null`, as here, or just return `PropertyError` from here?

alexsnaps (commented):
This is making this a whole lot better... thanks 🙏

src/auth_action.rs
```diff
@@ -196,13 +197,15 @@ mod test {
...
-        assert!(auth_action.conditions_apply());
+        assert!(auth_action.conditions_apply().is_ok());
+        assert!(auth_action.conditions_apply().expect("is ok"));
     }
 
     #[test]
     fn when_all_predicates_are_truthy_action_apply() {
         let auth_action = build_auth_action_with_predicates(vec!["true".into(), "true".into()]);
-        assert!(auth_action.conditions_apply());
+        assert!(auth_action.conditions_apply().is_ok());
```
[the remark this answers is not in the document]
adam-cattermole: Sure yeah I see your point, even though I do `expect` it on the next line it'll not reach that in err case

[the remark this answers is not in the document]
adam-cattermole: Yeah I was agreeing with you when I said this:
> even though I do expect it on the next line it'll not reach that in err case

Anyway I have fixed this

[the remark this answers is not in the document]
alexsnaps: Sorry, wasn't sure... that sentence confused me. My bad

alexsnaps (approved):

alexsnaps (approved):

Issue #181, opened 2025-04-24
wasm-shim: Extend rate limiting impl to allow setting hits_addend to a specific number


david-martin:
- Give a means to specify a counter increment amount (currently, [hard-coded](https://github.com/Kuadrant/wasm-shim/blob/main/src/service/rate_limit.rs#L18) to `1`)

PR #183, opened 2025-05-29
allow setting custom ratelimit domain and hits_addend values

Patryk-Stefanski:
allows setting custom ratelimit.hits_addend and ratelimit.domain values through actions.data field. This will be used for calls to limitador

Fix https://github.com/Kuadrant/wasm-shim/issues/181
Related to: https://github.com/Kuadrant/architecture/pull/124

### DIFF
--- a/src/ratelimit_action.rs
+++ b/src/ratelimit_action.rs
@@ -11,6 +11,7 @@ use cel_interpreter::Value;
 use cel_parser::ParseError;
 use log::{debug, error};
 use protobuf::RepeatedField;
+use std::collections::HashMap;
 use std::rc::Rc;
 
 #[derive(Debug)]
@@ -18,6 +19,8 @@ struct DescriptorEntryBuilder {
     pub key: String,
     pub expression: Expression,
 }
+
+pub const KNOWN_ATTRIBUTES: [&str; 2] = ["ratelimit.domain", "ratelimit.hits_addend"];
 
 impl DescriptorEntryBuilder {
     pub fn new(data_type: &DataType) -> Result<Self, ParseError> {
@@ -109,7 +112,9 @@ impl ConditionalData {
 
         let mut entries = RepeatedField::default();
         for entry_builder in self.data.iter() {
-            entries.push(entry_builder.evaluate()?);
+            if !KNOWN_ATTRIBUTES.contains(&entry_builder.key.as_str()) {
+                entries.push(entry_builder.evaluate()?);
+            }
         }
 
         Ok(entries)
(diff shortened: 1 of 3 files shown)


src/runtime_action.rs
```diff
@@ -155,13 +155,25 @@ impl RuntimeAction {
         match self {
             RuntimeAction::RateLimit(rl_action) => {
                 let descriptor = rl_action.build_descriptor()?;
-                if descriptor.entries.is_empty() {
+                let known_attrs = rl_action.get_known_attributes().unwrap_or_default();
+
+                if descriptor.entries.is_empty() && known_attrs.is_empty() {
```
alexsnaps, line 160: I don't think we want to do the call when `known_attrs` are there... afaik we _need_ at least one descriptor entry for limitador to be able to discriminate which `Limit`s to apply

src/ratelimit_action.rs
```diff
@@ -141,11 +142,40 @@ impl RateLimitAction {
             entries.extend(conditional_data.entries()?);
         }
 
+        let known_attributes: Vec<String> = vec![
+            "ratelimit.domain".to_string(),
+            "ratelimit.hits_addend".to_string(),
```
alexsnaps, line 147: Can we consolidate that vec? We're building it twice (again in `get_known_attributes`)... mostly concerned we don't keep them in sync

src/service/rate_limit.rs
```diff
@@ -47,7 +49,7 @@ mod tests {
         field.set_entries(RepeatedField::from_vec(vec![entry]));
         let descriptors = RepeatedField::from_vec(vec![field]);
 
-        RateLimitService::request_message(domain.to_string(), descriptors.clone())
+        RateLimitService::request_message(domain.to_string(), descriptors.clone(), 1)
```
alexsnaps, line 52: domain (which is a `String`?) `.to_string()`... why? what am I missing?

src/runtime_action.rs
eguzki: Right, the following action should not trigger gRPC call. 

```yaml
actions:
- service: ratelimit-service
  scope: ratelimit-scope-a
  data:
    - expression:
        key: ratelimit.domain
        value: foobar
    - expression:
        key: ratelimit.hits_addend
        value: '3'
```

src/ratelimit_action.rs
```diff
@@ -141,11 +142,40 @@ impl RateLimitAction {
...
         Ok(res)
     }
 
+    pub fn get_known_attributes(&self) -> Result<HashMap<String, String>, EvaluationError> {
+        let known_attributes = ["ratelimit.domain", "ratelimit.hits_addend"];
+        let mut result = HashMap::new();
+
+        for conditional_data in &self.conditional_data_sets {
+            for entry_builder in &conditional_data.data {
+                let entry = entry_builder.evaluate()?;
```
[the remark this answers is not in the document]
eguzki: Additionally, it is evaluating twice all the expressions, isn't it?

eguzki:
We should document this feature somewhere. Somebody from the control plane not familiar  with wasm-shim code should have a way to know how to override hits_addend or domain.

Patryk-Stefanski:
Closing in favour of https://github.com/Kuadrant/wasm-shim/pull/187 to pass license check.

PR #187, opened 2025-06-03
allow setting custom ratelimit domain and hits_addend values

Patryk-Stefanski:
allows setting custom ratelimit.hits_addend and ratelimit.domain values through actions.data field. This will be used for calls to limitador

Fix https://github.com/Kuadrant/wasm-shim/issues/181
Related to: https://github.com/Kuadrant/architecture/pull/124

### DIFF
--- a/src/ratelimit_action.rs
+++ b/src/ratelimit_action.rs
@@ -419,6 +419,16 @@ mod test {
             .collect::<RepeatedField<HeaderValue>>()
     }
 
+    fn build_action_for_known_attribute(key: &str, value: &str) -> Action {
+        let data = vec![DataItem {
+            item: DataType::Expression(ExpressionItem {
+                key: key.into(),
+                value: value.into(),
+            }),
+        }];
+        build_action(String::new(), vec!["true".into()], data)
+    }
+
     #[test]
     fn empty_predicates_do_apply() {
         let action = build_action(String::new(), Vec::default(), Vec::default());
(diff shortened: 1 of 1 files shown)


src/ratelimit_action.rs
```diff
@@ -19,6 +20,8 @@ struct DescriptorEntryBuilder {
     pub expression: Expression,
 }
 
+pub const KNOWN_ATTRIBUTES: [&str; 2] = ["ratelimit.domain", "ratelimit.hits_addend"];
```
alexsnaps, line 23: nit: no need for this to be `pub`lic afaict

src/runtime_action.rs
```diff
@@ -155,13 +155,25 @@ impl RuntimeAction {
...
                 let descriptor = rl_action.build_descriptor()?;
+                let known_attrs = rl_action.get_known_attributes()?;
+
                 if descriptor.entries.is_empty() {
                     debug!("build_message(rl): empty descriptors");
                     Ok(None)
                 } else {
+                    let domain = known_attrs
+                        .get("ratelimit.domain")
+                        .cloned()
```
alexsnaps, line 166: You should be able to "merge these 2 calls" into a single [`.remove`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.remove) and avoid the `clone()`ing of the string. Not a huge deal at all, but so you know :)

alexsnaps (approved):

src/ratelimit_action.rs
Patryk-Stefanski: done

alexsnaps:
> This configuration is the result of having one route match `GET /cars` and two limits, namely `limit_a` and `limit_b`. Two actions with the same scope and service. The merging process in place today, would generate duplicated known attributes.
> 
> Thoughts @alexsnaps @Patryk-Stefanski ??

Right, we discussed that... maybe during a call where you were not there. I guess you mostly mean conflicting entries, not really dups (as I guess that's really straightforward to fix... But, what if one says `3` while the other says `foo.bar * 3`... 
But, as far as I can tell, this isn't an issue with this bit of code, but rather an issue of the kuadrant operator pushing it to the wasm-shim. Let me explain: what does that mean at the policy level? That's the question that needs to be answered and be opinionated on _at the kuadrant operator level_, i.e. _how do we want to make that behave_? This is conflicting configuration _by the user_. Should we reject the "last" policy added as "contradicting"? Or what should happen? But I think, at this layer, everything is actually fine.

PR #189, opened 2025-06-12
Call out to limitador to check token usage limits

### LINKED ISSUE #179: wasm-shim: Call out to limitador to check token usage limits
This will be done at the request phase after sufficient descriptors are available
* model is parsed from the request body
* user info is set by authorino in the attributes
* other info, like request path, should already be available in envoy attributes

This implies we're parsing the request body at this phase to get the model (which we can then put into a header for later use if needed)

eguzki:
### What

Fixes #179 

TODO:
- [X] gRPC calls at request body phase
- [ ] `requestBodyJSON` CEL function


Cargo.lock
```diff
@@ -1,6 +1,6 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 3
```
eguzki, line 3: Is this change expected :shrug: ??

Cargo.toml
```diff
@@ -23,6 +23,7 @@ debug-host-behaviour = []
 [dependencies]
 proxy-wasm = { git = "https://github.com/Kuadrant/proxy-wasm-rust-sdk.git", branch = "kuadrant" }
 serde_json = "1.0"
+gjson = "0.8"
```
alexsnaps, line 26: Let's not use that. For one, we have `serde_json` already (just above), but beyond that, [`gjson`](https://crates.io/crates/gjson) looks pretty dead wrt to maintenance even.

src/data/attribute.rs
```diff
@@ -222,6 +222,19 @@ fn process_metadata(s: &Struct, prefix: String) -> Vec<(String, String)> {
     result
 }
 
+pub fn store_request_body_size(body_size: usize) -> Result<(), PropertyError> {
+    let attr = format!("{KUADRANT_NAMESPACE}\\.internal\\.request_body_size");
+    set_attribute(attr.as_str(), (body_size as u64).to_le_bytes().as_slice())
```
alexsnaps, line 227: Why storing that _outside_ of the `HttpContext`/`KuadrantFilter` ? Some comment somewhere implies it needs to be made available to the CEL's `Context` as well, which also confuses me...

Cargo.lock
alexsnaps: Ha! What toolchain are you using?

src/data/attribute.rs
eguzki: When `requestBodyJSON` is invoked (at CEL eval time), there  is no access to HttpContext/KuadrantFilter. This function needs to read request body using `proxy_wasm::hostcalls::get_buffer` and it needs the body size as input arg. [here](https://github.com/Kuadrant/wasm-shim/pull/189/files#diff-5c9e9bf6e990ebde97176b69053f106a454e54ea378947a0e1faaf5e25a51580R115)

src/runtime_action.rs
```diff
@@ -173,6 +179,50 @@ impl RuntimeAction {
     }
 }
 
+fn minimum_required_phase(fn_names: Vec<&str>) -> Phase {
```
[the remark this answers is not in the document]
eguzki: >  Can you get to request headers from the request body phase, or worse the response phase?

I am afraid, you can't. It's "minimum" in the sense that the gRPC request cannot be send before that phase. However, you are right that within an action, some expressions need to be evaluated at request headers phase and others at request body phase. That means that we might need to keep a (descriptor entry) structure that is being built at different phases. Makes sense?

src/data/attribute.rs
alexsnaps: Just flip the indirection on its head. Have the size populated from the call site

src/runtime_action.rs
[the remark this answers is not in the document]
alexsnaps: Right, I think it's not "minimal" but _the phase at which the expression can be evaluated_. We'd need some "lazy" concept and evaluate when possible. No need to keep the descriptor around tho. "Just" the expressions. This isn't specific to RL (and as such descriptors), you could require the same behavior for another service.

src/data/attribute.rs
alexsnaps: Today you're sort of using the `metadata` storage as a "request local static" to store the body size when it's made available, then retrieve it back when you need it. But you don't need to store it on the filter's state, internal to the host, to achieve that. 

I have mentioned in the past that probably having something "representing the `Request`" is what we would need to keep that state around... and avoid going back to the host all the time, e.g. to retrieve the same header 5 times... That thing could just get the body size injected when the phase passes it to the filter (that in turn would hold the instance of that `Request`, shared with the CEL `Context` for evaluation)

eguzki: ok, that's clearer now. TBH I never liked this mechanism of using host metadata to keep data for later usage. I thought about injecting context, but the impact was way bigger than the "temporary storage" solution implemented here. Check our slack conversation on 2nd of June :wink: you might have now more context to understand what I mean at that time :smile:

PR #193, opened 2025-07-04
Addition of new cher rate limit action

Patryk-Stefanski:


src/service.rs
```diff
@@ -66,6 +66,11 @@ impl GrpcService {
                 name: RATELIMIT_SERVICE_NAME,
                 method: RATELIMIT_METHOD_NAME,
             },
+            ServiceType::RateLimitCheck => Self {
```
eguzki, line 69: This is the only effective addition from the new service. Can we reuse RateLimitAction as the "shared" handler for RateLimit and RateLimitCheck actions?

PR #196, opened 2025-07-14
kuadrant filter owns running the workflow

eguzki:
The Kuadrant filter has been refactored to take ownership of executing the action workflow.
This groundwork lays the foundation for simplifying the sharing of the request context with CEL expressions and resuming the workflow during the request body phase. The scope and lifecycle of the request context are bounded to the Kuadrant filter. This context must be passed to actions in order to evaluate their associated expressions. With the proposed changes, the Kuadrant filter can directly access the actions and pass along the shared request context. The workflow implementation has also been refactored to support execution during the request body phase, which will be necessary for upcoming work.

The `GrpcErrResponse` struct has been renamed. It was previously used to encapsulate all necessary data for returning a direct response. It has been replaced with a more natural and descriptive name, better reflecting its purpose: representing a direct response that includes a status code, response headers, and a body.

Previously, negative responses—such as 4xx status codes like Forbidden or Over Limit, returned by Limitador or Authorino, were treated as errors and wrapped in a GrpcErrResponse. This behavior has been changed, as it was considered incorrect. These responses are now properly treated as explicit instructions from external services to abort the request workflow and return a direct response.

This changes provide a loop that can be executed at any point, regardless of the current request phase. This loop should be callable during the request_headers phase, the grpc_response phase, and, in the future, during on_request_body, on_response_headers, and on_response_body. Unfortunately, the current find_next_grpc_request method cannot serve this purpose.

Moreover, identifying the next gRPC request should not be an implementation detail of gRPC response processing. I’ve decoupled these concerns to allow for the gRPC response to be processed first—handling any errors—and then to resume the action workflow loop.

I am also refining the definition of operation variants. Sending a gRPC request is not an operation that depends on prior processing; it is simply the natural next step when there are no errors in processing the previous response or constructing the grpc request. Waiting for a gRPC response is not an operation either—it's a passive state that follows once a request has been sent.

Thus, processing a gRPC response may lead to one of the following outcomes:

*  The action succeeded, and there are pending operations to execute at the appropriate time (e.g., adding request or response headers).

*  The action succeeded, but the workflow must be aborted due to the presence of a downstream direct response.

*  The action failed. In this case, the failure mode must be checked to determine whether to terminate or continue the workflow.

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -2,7 +2,6 @@ pub(crate) mod auth;
 pub(crate) mod rate_limit;
 
 use crate::configuration::{FailureMode, Service, ServiceType};
-use crate::envoy::DeniedHttpResponse;
 use crate::envoy::{HeaderValue, HeaderValueOption, StatusCode};
 use crate::service::auth::{AUTH_METHOD_NAME, AUTH_SERVICE_NAME};
 use crate::service::rate_limit::{RATELIMIT_METHOD_NAME, RATELIMIT_SERVICE_NAME};
@@ -237,21 +236,6 @@ impl DirectResponse {
 
     pub fn body(&self) -> &str {
         self.body.as_str()
-    }
-}
-
-impl From<DeniedHttpResponse> for DirectResponse {
-    fn from(resp: DeniedHttpResponse) -> Self {
-        let status_code = resp.get_status().get_code();
-        let response_headers = resp
-            .get_headers()
-            .iter()
-            .map(|header| {
-                let hv = header.get_header();
-                (hv.key.to_owned(), hv.value.to_owned())
-            })
-            .collect();
-        Self::new(status_code as u32, response_headers, resp.body)
     }
 }
 
--- a/src/service/auth.rs
+++ b/src/service/auth.rs
@@ -1,9 +1,10 @@
 use crate::data::{get_attribute, PropError, PropertyError};
 use crate::envoy::{
     Address, AttributeContext, AttributeContext_HttpRequest, AttributeContext_Peer,
-    AttributeContext_Request, CheckRequest, Metadata, SocketAddress,
+    AttributeContext_Request, CheckRequest, DeniedHttpResponse, Metadata, SocketAddress,
 };
 use crate::service::errors::BuildMessageError;
+use crate::service::DirectResponse;
 use chrono::{DateTime, FixedOffset};
 use protobuf::well_known_types::Timestamp;
 use protobuf::Message;
@@ -13,6 +14,21 @@ use std::collections::HashMap;
 
 pub const AUTH_SERVICE_NAME: &str = "envoy.service.auth.v3.Authorization";
 pub const AUTH_METHOD_NAME: &str = "Check";
+
+impl From<DeniedHttpResponse> for DirectResponse {
+    fn from(resp: DeniedHttpResponse) -> Self {
+        let status_code = resp.get_status().get_code();
+        let response_headers = resp
+            .get_headers()
+            .iter()
+            .map(|header| {
+                let hv = header.get_header();
+                (hv.key.to_owned(), hv.value.to_owned())
+            })
+            .collect();
+        Self::new(status_code as u32, response_headers, resp.body)
+    }
+}
 
 pub struct AuthService;
 


adam-cattermole (commented):
I'm trying to get through all of the changes here but this is a lot in a single commit.

From what I can gather though we've opted to move away from the previous design decision to have separation of concerns of the filter from the operations in favour of fewer abstractions and tighter coupling to the filter. We made this decision during the refactoring to move to this approach so I'm a little concerned about the large change - it's also hard as part of this large refactor to determine how every case has been affected, but I'll keep going

src/auth_action.rs
```diff
@@ -44,78 +46,57 @@ impl AuthAction {
...
+                    Err(ProcessGrpcMessageError::UnsupportedField)
                 } else if !ok_response.get_query_parameters_to_set().is_empty() {
                     warn!("process_response(auth): Unsupported field 'query_parameters_to_set' in OkHttpResponse");
-                    return Err(GrpcErrResponse::new_internal_server_error());
+                    Err(ProcessGrpcMessageError::UnsupportedField)
                 } else if !ok_response.get_query_parameters_to_remove().is_empty() {
                     warn!("process_response(auth): Unsupported field 'query_parameters_to_remove' in OkHttpResponse");
-                    return Err(GrpcErrResponse::new_internal_server_error());
+                    Err(ProcessGrpcMessageError::UnsupportedField)
+                } else {
```
adam-cattermole, line 81: Does this still enforce an error or does this take into account `FailureMode` now? these cases should all error regardless of `FailureMode`

eguzki: grpc response processing result is checked in `src/runtime_action.rs` for all type of actions. Currently, `FailureMode` is applied for `ProcessGrpcMessageError::UnsupportedField` and any other error. 

iiuc, currently, when processing the grpc response, errors from parsing the response bytes (ProtobufError) can be bypassed by FailureMode but the actual errors processing by rate limit action or auth action should all fail regardless of FailureMode. Is that correct? Can I ask what is the rationale for doing that? why differentiation?

I am happy to keep that behavior, it's two liner change. But I want to understand the differences :pray: .

adam-cattermole: I'm unsure exactly for the ratelimit action, I believe the behaviour was replicating the existing behaviour. For these cases in the auth action; the fields are unsupported by the external auth service and could be an indication of a man in the middle attack, so the request should not proceed

src/filter/operations.rs
```diff
@@ -1,107 +1,25 @@
...
+#[derive(Debug)]
 pub enum Operation {
-    SendGrpcRequest(GrpcMessageSenderOperation),
-    AwaitGrpcResponse(GrpcMessageReceiverOperation),
-    AddHeaders(HeaderOperation),
-    Die(GrpcErrResponse),
-    // Done indicates that we have no more operations and can resume the http request flow
-    Done(),
+    EventualOps(Vec<EventualOperation>),
+    DirectResponse(DirectResponse),
```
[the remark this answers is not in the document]
eguzki: State transitions are still being handled entirely within the filter. However, I need a loop that can be executed at any point, regardless of the current request phase. This loop should be callable during the request_headers phase, the grpc_response phase, and, in the future, during on_request_body, on_response_headers, and on_response_body. Unfortunately, the current `find_next_grpc_request` method cannot serve this purpose. 

Moreover, identifying the next gRPC request should not be an implementation detail of gRPC response processing. I’ve decoupled these concerns to allow for the gRPC response to be processed first—handling any errors—and then to resume the action workflow loop.

I am also refining the definition of operation variants. Sending a gRPC request is not an operation that depends on prior processing; it is simply the natural next step when there are no errors in processing the previous response or constructing the grpc request. Waiting for a gRPC response is not an operation either—it's a passive state that follows once a request has been sent.

Thus, processing a gRPC response may lead to one of the following outcomes:

 *   The action succeeded, and there are pending operations to execute at the appropriate time (e.g., adding request or response headers).

 *   The action succeeded, but the workflow must be aborted due to the presence of a downstream direct response.

  *  The action failed. In this case, the failure mode must be checked to determine whether to terminate or continue the workflow.

src/runtime_action.rs
```diff
@@ -131,37 +126,57 @@ impl RuntimeAction {
...
-            },
-            Self::RateLimit(rl_action) => match Message::parse_from_bytes(msg) {
-                Ok(rate_limit_response) => rl_action.process_response(rate_limit_response),
-                Err(e) => {
-                    debug!("process_response(rl): failed to parse response `{e:?}`");
-                    self.resolve_failure_mode()
+        let res = match self {
+            Self::Auth(auth_action) => {
+                let check_response =
+                    Message::parse_from_bytes(msg).map_err(ProcessGrpcMessageError::from)?;
```
alexsnaps, line 156: No need to `map_err` here, `?` will `From` (see [here](https://github.com/rust-lang/rust/issues/31436#issuecomment-299638299)) automagically.

src/runtime_action.rs
```diff
@@ -131,37 +126,57 @@ impl RuntimeAction {
...
+                auth_action.process_response(check_response)
+            }
+            Self::RateLimit(rl_action) => {
+                let rate_limit_response =
+                    Message::parse_from_bytes(msg).map_err(ProcessGrpcMessageError::from)?;
```
alexsnaps, line 161: Same: No need to `map_err` here, `?` will `From` (see [here](https://github.com/rust-lang/rust/issues/31436#issuecomment-299638299)) automagically.

src/service.rs
```diff
@@ -44,6 +46,20 @@ pub(super) mod errors {
...
         }
     }
+
+    #[derive(Debug)]
+    pub enum ProcessGrpcMessageError {
+        Protobuf(ProtobufError),
+        Property(PropertyError),
+        EmptyResponse,
+        UnsupportedField,
+    }
```
alexsnaps, line 56: We don't really do anything with the granularity... If this is a thing to be useful "later", it should be introduced then.
Right now, this could be simplified... and would simplify the `match`es that deals with the error.

src/service.rs
```diff
@@ -220,6 +240,21 @@ impl GrpcErrResponse {
     }
 }
 
+impl From<DeniedHttpResponse> for DirectResponse {
```
alexsnaps, line 243: I'd move this impl to the "auth" module. This (looking at the `use`s) is already a little too coupled to impl. details, which probably should all be cleanup too.

src/filter/kuadrant_filter.rs
```diff
@@ -1,22 +1,24 @@
...
-use crate::filter::operations::{
-    GrpcMessageReceiverOperation, GrpcMessageSenderOperation, Operation,
-};
+use crate::configuration::FailureMode;
+use crate::filter::operations::{EventualOperation, Operation};
+use crate::runtime_action::IndexedRequestResult;
 use crate::runtime_action_set::RuntimeActionSet;
-use crate::service::{GrpcErrResponse, GrpcRequest, HeaderResolver, Headers};
+use crate::service::errors::ProcessGrpcMessageError;
+use crate::service::{DirectResponse, GrpcRequest, HeaderResolver, Headers, IndexedGrpcRequest};
```
alexsnaps, line 7: There is quite some additional coupling added here... While I can see why the changes made make this required, I can't judge on the merits of them.

alexsnaps (approved):
There is an implicit ask "to blindly trust" the changes proposed. Coupling between modules as been increased, which is certainly not a good sign. Now it could be the modules are not reflecting the new emerging design here that clearly moves away of the work that has happened until now. 

I don't know if this is "temporary" and will get better with follow up PRs, as this actually is "just a refactoring" and a departure from design principles that had been set. Only seeing were this eventually goes will tell us... So approving this for now, as I think no other work is happening here and we can consider "`main`" a feature branch for a little. 

A few general remarks, some of which I have made in the past, but for some reason(s) have been discarded:

 - Please try to keep refactorings, pure refactorings: i.e. _no test changes_;
 - Try to isolate individual changes in small changesets (commits);
 - Providing clear rational for design changes and how some previous goals aren't desirable anymore and why the new ones are superior, would be really appreciated, regardless of whether the PR contains the actual usage of the changes or not (like in this case, which makes it all really guess work).
 - Probably better, bring these changes up prior to opening the PR with others; tho I can also see how the design emerged as the changes were made.

eguzki:
> There is an implicit ask "to blindly trust" the changes proposed

There is no such ask. I tried to explain [here](https://github.com/Kuadrant/wasm-shim/pull/196#issue-3228919782) and [here](https://github.com/Kuadrant/wasm-shim/pull/196#discussion_r2209942230) which I did not hear any constructive comment yet

alexsnaps:
> This PR is the first one: `Make the change easy; (i.e. refactor/prepare the ground);`. The second one is sharing the request context (`Make the actual easy change (actually make use of the proposed changes)`)

That's a communication problem then, I was asking for these to be part of the _same_ PR, but in 2 separate commits. 

wrt squashing, if a change is "irrelevant", now that it's squashed it gets harder to "ignore". That is what I was trying to convey: keep related changes together in individual commits, e.g. that "remove the RC", was actually great! It either lets people review changes in isolation and hopefully tells a story of how the all comes together (e.g. make the change easy changeset; then the make the easy change one).

eguzki:
> You really think [this commit](https://github.com/Kuadrant/wasm-shim/pull/195/commits/121fb938ef596ba167f2d9ceab6bfb9f546acfb5) _only_ "`Make the change easy; (i.e. refactor/prepare the ground)`", do you?

Fair enough. This commit is groundwork not only for sharing context during CEL evaluation, but also for broader changes, and I failed to communicate that clearly.

The commit lays the foundation for executing actions during the request and response body phases. As mentioned earlier, I need a loop that can be executed at any point, regardless of the current request phase. This loop should be callable during the request_headers phase, the grpc_response phase, and, in the future, during on_request_body, on_response_headers, and on_response_body. Unfortunately, the current `find_next_grpc_request` method cannot fulfill this need.

src/auth_action.rs
eguzki: @adam-cattermole I added a [commit](https://github.com/Kuadrant/wasm-shim/pull/196/commits/bb77ba497d3caed9f77a5e8fec3ebad99bd8bc6c) addressing the comment. Hopefully it is like you asked.

src/runtime_action.rs
eguzki: nice! comment addressed

src/service.rs
eguzki: It turns out that to address this [comment](https://github.com/Kuadrant/wasm-shim/pull/196#discussion_r2209835383), `UnsupportedField` needs to be used ( auth action unsupportedfield errors stop flow regardless of FailureMore). Will remove `EmptyResponse` and added when needed.

PR #199, opened 2025-07-17
gRPC calls at response body phase & responseBodyJSON CEL function

eguzki:
### What 

Fixes https://github.com/Kuadrant/wasm-shim/issues/182

Introduces `responseBodyJSON( ... )` CEL function. 

Addtional notes: 
* On rate limiting calls at response body phase, the returned headers cannot be added to the downstream response, as it is too late to do so. 
* `response.*` attributes are not pre-fetched. So, basically, still not supported (like currently). 

## `responseBodyJSON(json_pointer)`

This CEL function parses response body as json and looks up a value by a JSON Pointer.
JSON Pointer defines a string syntax for identifying a specific value within a JavaScript Object Notation (JSON) document.
A Pointer is a Unicode string with the reference tokens separated by `/`.
For more information read [RFC6901](https://datatracker.ietf.org/doc/html/rfc6901).

Example:

when the response body is:

```json
{
  "my": {
    "value": "hello",
    "list": ["a", "b", "c"]
  }
}
```
and the expression is:

```yaml
data:
- expression:
    key: my_value
    value: responseBodyJSON('/my/value')
```

it evaluates to: `"hello"` CEL value. Similarly,

`responseBodyJSON('/my/list/1')` evaluates to `"b"` CEL value.

`responseBodyJSON('/a/b/c')` evaluates to `Null` CEL value.

### Verification steps

* Compile the wasm module

```sh
make build
```

* Run `examples/response_body`

This configuration parses the response body as a JSON serialized string and picks the `/usage/total_tokens` key whose value is sent to the rate limiting server.

* change directory
```sh
cd examples/response_body
```
* Run the environment
```sh
make run
```

* Check model based rate limiting works
```
curl --resolve body-request.example.com:18000:127.0.0.1 "http://body-request.example.com:18000"/v1/chat/completions \
  -d '{
    "message": [
      {
        "role": "user",
      }
    ]
  }'
```
It should return `200 OK`.

Expected rate limiting service logs:

```sh
docker compose logs -f rlsbin
```

```console
rlsbin-1  | 2025-07-17T16:43:00.062Z DEBUG [rlsbin::envoy_rls::server] Request received: Request { metadata: MetadataMap { headers: {"te": "trailers", "grpc-timeout": "20m", "content-type": "application/grpc", "x-envoy-internal": "true", "x-envoy-expected-rq-timeout-ms": "20"} }, message: RateLimitRequest { domain: "domain-a", descriptors: [RateLimitDescriptor { entries: [Entry { key: "tokens", value: "24" }], limit: None }], hits_addend: 1 }, extensions: Extensions }
```
which contains the desired descriptor entry `entries: [Entry { key: "tokens", value: "24" }]`.

> Note: the tokens value may be different

### Clean up

```
make clean
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ It can also be used in predicates:
 
 ```yaml
 predicates:
-- requestBodyJSON('/my/value')
+- requestBodyJSON('/my/value') == 'hello'
 ```
 
 #### `responseBodyJSON(json_pointer)`
@@ -141,7 +141,7 @@ It can also be used in predicates:
 
 ```yaml
 predicates:
-- responseBodyJSON('/my/value')
+- responseBodyJSON('/my/value') == 'hello'
 ```
 
 ### Well Known Attributes


Patryk-Stefanski:
Verification steps were successful and code looks good to me. /lgtm

alexsnaps (dismissed):

README.md
```diff
@@ -98,6 +98,52 @@ predicates:
...
+`responseBodyJSON('/my/list/1')` evaluates to `"b"` CEL value.
+
+`responseBodyJSON('/a/b/c')` evaluates to `Null` CEL value.
+
+
+It can also be used in predicates:
+
+```yaml
+predicates:
+- responseBodyJSON('/my/value')
```
Patryk-Stefanski, line 144:
```suggestion
- responseBodyJSON('/my/value')=="hello"
```

Patryk-Stefanski: either way we have to compare the predicate to something, and in the example [here](https://github.com/Kuadrant/wasm-shim/pull/199/files/f3ca3c1af8761b8da28b3081f23d499d2313a11c#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R119) we use "hello"

PR #204, opened 2025-08-06
Address coupling issues

eguzki:
## What 

This pull request is a pure refactoring effort to address architectural concerns (noted in #198, #199) regarding tight coupling in our gRPC call implementation. It introduces no new features or bug fixes but improves code quality, maintainability, and extensibility.

Key Improvements

### Decoupling via "Transient Attributes"

The data module is no longer aware of specific request/response phases (e.g., requestBody, responseBody).

* Problem: The previous design coupled the core evaluation logic to the HTTP request lifecycle.

* Solution: We've introduced the concept of _transient_ attributes. These are attributes that may not be available upon the first evaluation attempt but are expected to appear later.

* Mechanism:
  * The CEL evaluator now detects required transient attributes (like requestBodyJSON) at parse time.

  * If an attribute is missing during evaluation, it returns a new, specific EvaluationError::TransientError.

  * The calling filter interprets this error as a signal to "wait" for more data (e.g., the request body) and then retry the evaluation.

  * This decouples the data module, which now only needs to know about a generic map of transient data.

### Structured Error Handling

Error handling is now more robust and informative.

* The CEL expression  _eval_ method now returns a structured `EvaluationError` instead of a simple `CelError`.

* This new error type allows callers to programmatically inspect the error's nature (e.g., `TransientError`).

 * `CelError` is now kept as a private implementation detail within the _data_ module, creating a cleaner public API.

### Testing Enhancements

The test suite has been strengthened to support and verify these changes.

* TestResolver has been refactored to be stricter, verifying call expectations precisely.

* Unit tests were updated to align with new function signatures (e.g., the new EvaluationError).

* Importantly, no changes were needed for the E2E tests, confirming that the external behaviour of the system remains unchanged.

Related Issues: Addresses concerns from #198, #199, #190.

### DIFF
--- a/src/data/attribute.rs
+++ b/src/data/attribute.rs
@@ -11,7 +11,7 @@ pub(super) mod errors {
     use std::error::Error;
     use std::fmt::{Debug, Display, Formatter};
 
-    #[derive(Debug, PartialEq)]
+    #[derive(Debug)]
     pub enum PropertyError {
         Get(PropError),
         Parse(PropError),
@@ -39,7 +39,7 @@ pub(super) mod errors {
         }
     }
 
-    #[derive(Debug, PartialEq)]
+    #[derive(Debug)]
     pub struct PropError {
         message: String,
     }
--- a/src/data/cel.rs
+++ b/src/data/cel.rs
@@ -32,14 +32,6 @@ pub(super) mod errors {
         got: ValueType,
     }
 
-    impl PartialEq for PredicateResultError {
-        fn eq(&self, other: &Self) -> bool {
-            // PartialEq not implemented by ValueType,
-            // thus, comparing out of the Display output 🤷
-            format!("{}", self.got) == format!("{}", other.got)
-        }
-    }
-
     impl Display for PredicateResultError {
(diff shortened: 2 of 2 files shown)


src/filter/kuadrant_filter.rs
```diff
@@ -29,8 +29,8 @@ pub(crate) struct KuadrantFilter {
     header_resolver: HeaderResolver,
     path_store: PathCache,
     grpc_message_receiver: Option<EventReceiver>,
-    request_body_receiver: Option<EventReceiver>,
-    response_body_receiver: Option<EventReceiver>,
+    request_body_receiver: Option<(EventReceiver, String)>,
```
eguzki, line 32: Is it worth declaring a new type?

```rust
type TransientAttr = String;
```
or 

```rust
struct TransientAttr(String);
```

alexsnaps (approved):
As a heads up, we plan on having a deep refactoring of this code base next week. 
Thanks for these changes in the meantime tho!

eguzki:
> As a heads up, we plan on having a deep refactoring of this code base next week. Thanks for these changes in the meantime tho!

Sure! Please keep me informed; I'm happy to contribute to the discussion and implementation.

PR #207, opened 2025-08-07
Use new conditional data struct from config

Patryk-Stefanski:
Makes use of the new conditional data structure added in [#1387](https://github.com/Kuadrant/kuadrant-operator/pull/1387)

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,14 +31,17 @@ actionSets:
     actions:
     - service: auth-service
       scope: auth-scope-a
+      predicates:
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -33,12 +33,13 @@ actionSets:
       scope: auth-scope-a
```
eguzki, line 33: Could you add a predicate here to highlight that predicates can be used as action top level attr?

eguzki (commented):
Looking good

PR #212, opened 2025-09-09
Enabling streaming for TRLP

Patryk-Stefanski:


src/filter/kuadrant_filter.rs
```diff
@@ -600,6 +631,271 @@ impl KuadrantFilter {
...
+        }
+    }
+    if out.is_empty() {
+        None
+    } else {
+        Some(out)
+    }
+}
+
+fn split_last_two_frames(buffer: &mut String) -> (Option<String>, Option<String>) {
```
eguzki, line 771: I implemented this unittests that I would expect to pass, but it is failing. 

```rust
   #[test]
    fn split_last_two_frames_with_partial_blocks() {
        let mut buf = String::from("data: {\"id");
        let (usage_frame, done_frame) = split_last_two_frames(&mut buf);
        assert!(usage_frame.is_none());
        assert!(done_frame.is_none());
        assert_eq!(buf, "data: {\"id");

        buf.push_str("\":1}\n\ndata: {");
        let (usage_frame, done_frame) = split_last_two_frames(&mut buf);
        assert!(usage_frame.is_none());
        assert_eq!(done_frame.unwrap(), "data: {\"id\":1}");
        assert_eq!(buf, "data: {\"id\":1}\n\ndata: {");

        buf.push_str("\"usage\":1}data: [DONE]\n\n");
        let (usage_frame, done_frame) = split_last_two_frames(&mut buf);
        assert_eq!(usage_frame.unwrap(), "data: {\"usage\":1}");
        assert_eq!(done_frame.unwrap(), "data: [DONE]");
        assert_eq!(buf, "data: [DONE]\n\n");
    }
```

Would you expect this test to pass?

eguzki (commented):
If I read this correctly, this is buffering the body until usage is found. Provided that the usage will be found by the end of the body, this is buffering almost the entire body. IF my reading is correct, could we optimize a bit taking advantage of the streaming nature in terms of blocks of the response body?

src/filter/kuadrant_filter.rs
```diff
@@ -600,6 +631,271 @@ impl KuadrantFilter {
...
             request_headers_to_add: Some(Vec::default()),
             phase: Phase::RequestHeaders,
+            response_content_type: None,
+            sse_buffer: String::new(),
+            sse_usage: String::new(),
+            sse_offset: 0,
+        }
+    }
+
+    fn handle_sse_stream(
```
eguzki, line 641: Maybe not for this PR. But I would have implemented this `handle_sse_stream` in a generic way, expecting generic Server Sent Events. This method would produce events and some other function may use those events as source for finding `usage` data.

src/filter/kuadrant_filter.rs
eguzki: > Each individual chunk should contain a complete, parseable JSON object.

That's an interesting point. In my experience, data streams often send chunks of arbitrary size, meaning a single JSON object could be split across multiple chunks. Is there a specific protocol or library you're thinking of that guarantees a complete object per chunk? I'd love to learn more about it if you have a link or some documentation.

Patryk-Stefanski: SSE events should be produced as complete json, I havent seen the wasm buffering chunks mid stream they always start and end with a complete json. 

In the event that the event is split between two chunks the logic shoudl cover this case by joining the buffer.

eguzki (approved):
This is working software, and as such, it now adds support for the SSE streaming format for OpenAI workloads 🥳

I strongly recommend two enhancements, which can be implemented in follow-up PRs:

* Memory management enhancement: Take advantage of the stream processing model by keeping only the necessary byte chunks in memory (e.g., the last two events or messages). When an `end_of_stream` signal is received, emit the usage data. Currently, (almost) the entire request body is buffered inside the WASM VM because the usage section appears at the end of the request body, just before the _END_ frame.
* Decouple SSE event processing: Move SSE event processing into a separate module that the Kuadrant filter can leverage. As a suggestion, this module can be stateful (buffering the necessary chunks) and expose a single public function:
```rust
fn process_sse_chunk(&self, chunk: &[u8]) -> Option<String>;
```

PR #221, opened 2025-09-26
delay ratelimit action when we need to eval requestBody

Patryk-Stefanski:


tests/request_body_transient_error.rs
```diff
@@ -0,0 +1,385 @@
...
+
+    let root_context = 1;
+    // Configuration with request_data that mixes transient (requestBodyJSON) and
+    // non-transient (request.method) expressions - this tests that the action waits
+    // for transient data while non-transient data can be evaluated immediately
+    let cfg = r#"{
+        "requestData": {
+            "metrics.labels.model": "requestBodyJSON('/model')",
+            "metrics.labels.max_tokens": "requestBodyJSON('/max_tokens')",
+            "metrics.labels.method": "request.method"
```
eguzki, line 226: This request Data expression `request.method` evaluates correctly because it's in the predicates section as well. If you remove `request-.method` from predicates, this should fail, as evaluating `request.method` at request body phase cannot be done. I will be fixing that in a separated PR

src/ratelimit_action.rs
```diff
@@ -208,7 +207,14 @@ impl RateLimitAction {
...
                         }
-                    });
+                        Err(e) if e.is_transient() => {
+                            //only interested in returning transient errors for so we can evaluate again later
+                            info!("build_message(rl): transient evaluation error for requestData field '{}': {}", field, e);
+                            return Err(BuildMessageError::Evaluation(Box::new(e)));
+                        }
+                        _ => {}
+                    }
+                }
```
adam-cattermole, line 217: As far as I understand this should not be done and is not safe - unless something changed, you cannot delay the request until the requestBody phase as it'll stream the headers to the upstream cc @alexsnaps 

See this document on stop iteration: https://docs.google.com/document/d/1Whd1C0k-H2NHrPOmlAqqauFz6ObSTP017juJIYyciB0/edit?tab=t.0#heading=h.17wg41voij6q

eguzki: Thats' why we pre-fetch the request.* attributes in advance.

eguzki: Now I undestand your initial statement about "you cannot delay the request until the requestBody phase as it'll already have sent the headers to the upstream". If that would be true, we would not be able to rate limit based on anything from the request body. Right? At the very least, we know for sure that request body will not be routed to the upstream service.

eguzki:
For future readers, to clarify, we are discussing the feasibility of applying rate limiting at the request body phase.

If we decide not to allow this, the call to the rate limit service will always occur at the request headers phase. In that case, we could still add metric labels derived from the request body, but those labels would not be available for authorized_calls and limited_calls, which are only incremented at the request headers phase.

If we decide to allow (and support) this, the call to the rate limit service would occur at the request body phase (when the user requires metric labels from the request body). This raises concerns about the gateway’s memory management, since the gateway would need to buffer the request headers. When the request headers exceed the maximum buffer size, the gateway could behave in one of two ways:

* It flushes the request headers to the upstream service, in which case the upstream service would not be fully secured.

* It raises an OutOfMemory error, in which case the gateway could panic, with unpredictable consequences. We have already observed cases where VM workers stopped when WASM code panicked.

PR #227, opened 2025-10-02
Dont pause stream when looking for usage with streaming

Patryk-Stefanski:


tests/streaming.rs
```diff
@@ -185,32 +183,6 @@ fn it_processes_usage_event_across_chunks_until_done() {
             Some(LogLevel::Debug),
             Some("#2 handle_stream: processing chunk: data: [DONE]\n\n"),
         )
-        .execute_and_expect(ReturnType::Action(Action::Pause))
-        .unwrap();
-
-    // Third call: end_of_stream true → should now send gRPC using captured usage
```
eguzki, line 191: Why removing this? Perhaps related to `wasm log: proxy_on_grpc_receive invalid context_id, no match found     thread=36` ?

tests/streaming.rs
```diff
@@ -133,84 +133,277 @@ fn it_processes_usage_event_across_chunks_until_done() {
...
+            Some(LogLevel::Debug),
+            Some("process_response(rl): received OK response"),
+        )
+        .execute_and_expect(ReturnType::None)
+        .unwrap();
+}
+
+#[test]
+#[serial]
+fn it_streams_chunks_without_pausing_until_end_of_stream() {
```
eguzki, line 198: Cool test!

eguzki (approved):
LGTM

well covered by tests

In short, the behavior is 
* wait till the end of the stream.
* Then, return what self.run() returns. That is,
  * if there is a grpc call to be done-> pause
	*On the grpc_response callback run self.resume_http_response() when no more actions are to be run.
  * if there is nothing to do -> continue

PR #234, opened 2025-10-10
Attribute cache and module structure

adam-cattermole:
* I've refactored the module structure for v2 into the following:

```
src/v2/
├── data
│   ├── attribute.rs
│   ├── cel
│   │   └── strings.rs
│   ├── cel.rs
│   └── mod.rs
├── kuadrant
│   ├── cache.rs
│   ├── context.rs
│   ├── mod.rs
│   └── resolver
│       ├── mock.rs
│       ├── mod.rs
│       └── wasm_host.rs
├── mod.rs
├── services
│   ├── auth.rs
│   └── mod.rs
└── temp
    └── mod.rs
```

* `/data` somewhat matches v1
* `/services` also somewhat matches v1, the content for the specific grpc service implementations
* `/kuadrant` currently contains the cache, and the `ReqRespCtx`
* `/resolver` has the `AttributeResolver` and it's implementations (mock + wasm) for retrieving from the host

Not tied to any of this, happy to change it, but we were in need of some structure

I've also made a couple changes to retain the attributes from the host in a cache


alexsnaps:
One random question here, why is the `resolver` module sitting at the same level as `kuadrant`? Am I reading this right and the only (well, aside some other test code in `data`, but let's ignore that, and mostly _test code_ for now) consumer is `kuadrant`?

PR #237, opened 2025-10-17
Bump dev claude command

eguzki:
Depends on https://github.com/Kuadrant/wasm-shim/pull/236. It should be merged first

### DIFF
--- a/.claude/commands/bump-dev.md
+++ b/.claude/commands/bump-dev.md
@@ -18,3 +18,4 @@ Important notes:
 - Bump the MINOR version (0.11.0-dev -> 0.12.0-dev, not 0.11.1-dev)
 - Update BOTH Cargo.toml and Cargo.lock (via cargo check)
 - Include "Signed-off-by" in commit message following project conventions
+- Include a comment in the commit message explicitly stating it was driven by an AI tool


Boomatang:
I have to ask the question why? There is so much energy required to bump one value. Is this really what we want. I took the command file your adding and gave it to claude to create a python script that does the same thing. It user the semver package to make it easier to bump the version with have to use regex. I have not tried it, but I can guess it would be more repeatable and testable than using an AI. The gist is attached below. 

I also had the though that cargo would have some thing for doing this. There is a cargo-edit package/module  that would bump the version how you want. There would need to be some scripting around generation the commit message, but really how hard would that be.

https://gist.github.com/Boomatang/2af4e9ed4f2e0e3e1f9a6b6994aa9588
https://github.com/killercup/cargo-edit?tab=readme-ov-file#cargo-set-version

didierofrivia (approved):
I'm approving this only because it was decided during the community call, and I know it's more about learning than anything... however, we can agree that this is far from ideal, it's an overkill, something that we could solve with a simple script and save who knows how many cycles of CPU somewhere... we should start considering the ethics and impact of our decisions as engineers, and care more about the hidden cost of using "AI" (resource hungry LLM hosted somewhere) for trivial things like these.

PR #242, opened 2025-10-20
Configuring Pipelines in `alternate_design`

adam-cattermole:


eguzki (approved):
let's merge this unfinished piece of work into `alternate-design`.

I like the Pipeline::Eval() function dealing with task_queue and deferred_task. I wonder what is the difference. When the action.apply response is Deferred, we could do the same as `Requeued`. Or what is the plan for the "deferred tasks"?

src/v2/kuadrant/pipeline/tasks/mod.rs
```diff
@@ -7,20 +7,21 @@ use crate::v2::services::Service;
...
-trait Task {
+pub trait Task {
     fn apply(self: Box<Self>, ctx: &mut ReqRespCtx) -> TaskOutcome;
 }
 
 #[allow(dead_code)]
-struct PendingTask {
+pub struct PendingTask {
     is_blocking: bool,
     allow_task: Option<Box<dyn Task>>,
```
adam-cattermole: This will more than likely become a vec of `allow_tasks` to perform, `deny_task` _might_ also but I haven't thought through all the scenarios yet

PR #245, opened 2025-10-23
Refactor handling of headers in tasks

adam-cattermole:
This builds on top of the `configuration` branch as I want to operate on tasks which now reside in `v2/kuadrant/pipeline/tasks`

The primary change here is updating the structure we were using to store headers in v2 from a `HashMap` to a `Vec` (same as the host), to properly retain duplicates. I've created a new `Headers` type that encapsulates the operations on `Headers` and provided means to convert


adam-cattermole:
Just moved the headers task to it's own submodule within common so that we can separate other common tasks; let me know if you agree with this change

PR #247, opened 2025-10-28
Task: parse AI token usage from response with SSE format

### LINKED ISSUE #240: How do we want to handle response body, 'streaming' like API?

eguzki:
Fixes #240


didierofrivia (approved):
So much was going on with the parser and SSE, I like it that now it's encapsulated. We could definitely be tuned for any parsing lib that we could replace in the future. 👍🏼

PR #248, opened 2025-10-28
Retain completed_tasks and include explicit dependence

adam-cattermole:
This is an attempt at implementing the explicit dependence between tasks.

- When generating the pipeline for now we'll include a dependency of each task prior to the current one; we could just increment sequentially where id = i and dependency is populated with i-1
- The tasks all implement the `dependencies_met` trait to decide whether they can proceed or not

My hope is this lets us move toward enabling this in the configuration with minimal changes - the actions will eventually have an ID, and a dependency field that shows what they are dependent on so we can plug that into our pipeline build logic:

```yaml
actions:
- id: A
  dependencies: [] # would be omitted as empty
  service: limitador
  scope: RLS-domain
  conditionalData:
  - data:
    - static:
        key: admin
        value: '1'
- id: B
  dependencies:
  - A
  service: authorino
  scope: authconfig-A
```

This would let us do the following:

- The AuthTask could implement `dependencies_met` as `true`, _but_ the task that actually sends the message could be false until the `completed_tasks` contains `A`
- This would let `AuthTask` transition to `SendTask` (or whatever we call it) (i.e. fetch data; craft `CheckRequest` and prepare to send) _but_ the sending itself would be blocked until ratelimiting task A completes

This means that to transition to a SendTask it must have filled the data dependency; once it has, to actually send it has to take into account the configuration dependence. This lets us eagerly evaluate everything we can until the point at which we are waiting to send; we then 'just' wait for our dependent tasks to finish before we fire the message.

---
Eventually once we have this in configuration we can verify the DAG makes sense or reject

### DIFF
--- a/src/v2/kuadrant/pipeline/executor.rs
+++ b/src/v2/kuadrant/pipeline/executor.rs
@@ -4,12 +4,13 @@ use crate::v2::kuadrant::{
     pipeline::tasks::{PendingTask, Task, TaskOutcome},
     ReqRespCtx,
 };
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashSet};
 
 pub struct Pipeline {
     ctx: ReqRespCtx,
     task_queue: Vec<Box<dyn Task>>,
     deferred_tasks: BTreeMap<usize, PendingTask>,
+    completed_tasks: HashSet<String>,
 }
 
 impl Pipeline {
@@ -18,6 +19,7 @@ impl Pipeline {
             ctx,
             task_queue: Vec::new(),
             deferred_tasks: BTreeMap::new(),
+            completed_tasks: HashSet::new(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/v2/kuadrant/pipeline/executor.rs
```diff
@@ -4,12 +4,13 @@ use crate::v2::kuadrant::{
...
     ReqRespCtx,
 };
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashSet};
 
 pub struct Pipeline {
     ctx: ReqRespCtx,
     task_queue: Vec<Box<dyn Task>>,
     deferred_tasks: BTreeMap<usize, PendingTask>,
+    completed_tasks: HashSet<String>,
```
didierofrivia: Would it make sense to create a type for the TaskID (String) ?

PR #250, opened 2025-10-29
RateLimit Task

didierofrivia:
This PR introduces the RateLimit Task, the `TaskOutcome` is not complete, needs to define the _deferred_ tasks to trigger.

Next PR: DirectResponse task + specific tasks for the deferred outcome

### DIFF
--- a/src/v2/kuadrant/pipeline/tasks/ratelimit.rs
+++ b/src/v2/kuadrant/pipeline/tasks/ratelimit.rs
@@ -66,7 +66,7 @@ const KNOWN_ATTRIBUTES: [&str; 2] = ["ratelimit.domain", "ratelimit.hits_addend"
 pub struct RateLimitTask {
     task_id: String,
     dependencies: Vec<String>,
-    is_blocking: bool,
+    pauses_filter: bool,
 
     // Rate limit configuration
     scope: String,
@@ -93,7 +93,7 @@ impl RateLimitTask {
         Self {
             task_id,
             dependencies,
-            is_blocking: true,
+            pauses_filter: true,
             scope,
             service,
             predicates,
@@ -213,9 +213,9 @@ impl RateLimitTask {
         ctx: &ReqRespCtx,
     ) -> Result<Vec<Entry>, EvaluationError> {
         if conditional_data.predicates.is_empty()
-            || conditional_data.predicates.iter().any(|expr| {
-                expr.eval(ctx)
-                    .is_ok_and(|v| matches!(v, AttributeState::Available(Value::Bool(true))))
+            || conditional_data.predicates.iter().any(|pred| {
+                pred.test(ctx)
+                    .is_ok_and(|v| matches!(v, AttributeState::Available(true)))
             })
         {
             conditional_data
@@ -239,7 +239,7 @@ impl Task for RateLimitTask {
             Ok(AttributeState::Available(descriptors)) => descriptors,
             Ok(AttributeState::Pending) => {
                 // Need to wait for attributes, requeue
-                return TaskOutcome::Requeued(self);
+                return TaskOutcome::Requeued(vec![self]);
             }
             Err(_e) => {
                 // TODO: Handle error appropriately based on failure mode
@@ -284,14 +284,13 @@ impl Task for RateLimitTask {
             token_id,
             pending: PendingTask {
                 task_id: Some(self.task_id),
-                is_blocking: self.is_blocking,
-                process_response: Box::new(move |response| {
-                    // Parse the rate limit response
-                    let rate_limit_response = match service.parse_message(response) {
+                pauses_filter: self.pauses_filter,
+                process_response: Box::new(move |ctx, _status_code, response_size| {
+                    let rate_limit_response = match service.get_response(ctx, response_size) {
                         Ok(parsed) => parsed,
                         Err(_e) => {
-                            // TODO: Handle parsing error and/or FailureMode task?
-                            return Vec::new();
+                            // TODO: Handle parsing error
+                            return TaskOutcome::Failed;
                         }
                     };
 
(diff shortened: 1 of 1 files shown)


src/v2/kuadrant/pipeline/tasks/ratelimit.rs
```diff
@@ -1,30 +1,344 @@
...
 
-        TaskOutcome::Done
+    /// Builds the rate limit request message from the context
+    fn build_descriptors(&self, ctx: &ReqRespCtx) -> BuildDescriptorsResult {
+        // Build descriptor entries by evaluating conditional data
+        let mut entries = Vec::new();
+
+        // TODO: Candidate for a `prepare` task/method
+        for conditional_data in &self.conditional_data_sets {
+            match self.predicates_apply(ctx) {
```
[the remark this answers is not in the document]
eguzki: Yes, that `self.predicates_apply` checks action predicates. Hence my question, are conditional_data's predicates being checked?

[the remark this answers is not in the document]
didierofrivia: After our call and you pointing out the correct way of building descriptors, I've applied the following changes: https://github.com/Kuadrant/wasm-shim/pull/250/commits/cd47d8ec1a004229a536a1baf52c6f9c749eba35

src/v2/kuadrant/pipeline/tasks/ratelimit.rs
```diff
@@ -1,36 +1,767 @@
...
+
+        for predicate in &self.predicates {
+            match predicate.test(ctx)? {
+                AttributeState::Available(true) => continue,
+                AttributeState::Available(false) => return Ok(AttributeState::Available(false)),
+                AttributeState::Pending => return Ok(AttributeState::Pending),
+            }
+        }
+        Ok(AttributeState::Available(true))
+    }
```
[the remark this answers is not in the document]
didierofrivia: Will refactor to avoid re calling and matching on result

src/v2/kuadrant/pipeline/tasks/ratelimit.rs
```diff
@@ -1,36 +1,785 @@
...
+            ))),
+        }
+    }
+}
+
+/// Type indicating the state of the descriptors build outcome
+enum BuildDescriptorsState {
+    Ready(Vec<RateLimitDescriptor>),
+    Pending,
+}
```
adam-cattermole: I've not reasoned out the behaviour of all of this in my head but this slightly concerns me; I _thought_ this would be handled by the current changes in the CEL library and the existing caching logic.

If it requires data at a phase it's not currently in, just 'trying' to build the descriptor is enough for it to cache things; if it then builds the descriptor later in the correct phase all of the data it needs should be there.

src/v2/kuadrant/pipeline/tasks/ratelimit.rs
```diff
@@ -1,36 +1,766 @@
...
+        dependencies: Vec<String>,
+        service: Rc<RateLimitService>,
+        scope: String,
+        predicates: Vec<Predicate>,
+        conditional_data_sets: Vec<ConditionalData>,
+    ) -> Self {
+        Self {
+            task_id,
+            dependencies,
+            pauses_filter: true,
```
adam-cattermole: I think this isn't necessarily always true for RL - it'll be false for the one that runs in response phase; but we can update this when we wire it into the blueprint

src/v2/kuadrant/pipeline/tasks/ratelimit.rs
```diff
@@ -1,36 +1,766 @@
...
+
+    /// Builds the rate limit descriptors from the context
+    fn build_descriptors(
+        &self,
+        ctx: &ReqRespCtx,
+    ) -> Result<AttributeState<Vec<RateLimitDescriptor>>, EvaluationError> {
+        // TODO: Candidate for a `prepare` task/method
+        if !self.predicates.is_empty() {
+            for predicate in &self.predicates {
+                match predicate.test(ctx)? {
```
adam-cattermole: Any reason why this logic is not using the exposed `predicates.apply()` method but iterates through them all? It's nice to have that encapsulated in one place and can match on the outcome here

src/v2/kuadrant/pipeline/tasks/ratelimit.rs
```diff
@@ -1,36 +1,766 @@
...
+    fn build_entries(
+        &self,
+        conditional_data: &ConditionalData,
+        ctx: &ReqRespCtx,
+    ) -> Result<Vec<Entry>, EvaluationError> {
+        if conditional_data.predicates.is_empty()
+            || conditional_data.predicates.iter().any(|pred| {
+                pred.test(ctx)
+                    .is_ok_and(|v| matches!(v, AttributeState::Available(true)))
+            })
```
adam-cattermole: I thought `all` the predicates had to be true not `any` to add these data items?

src/v2/kuadrant/pipeline/tasks/ratelimit.rs
```diff
@@ -1,36 +1,766 @@
...
-    allow_task: Option<Box<dyn Task>>,
-    deny_task: Box<dyn Task>,
+    service: Rc<RateLimitService>,
+
+    // Conditional data for building descriptors
+    conditional_data_sets: Vec<ConditionalData>,
+    predicates: Vec<Predicate>,
+
+    // Default hits addend
+    default_hits_addend: u32,
```
adam-cattermole: I wonder if this should be a const; I know it's overridden by the 'well known attributes' but I can't think of a reason the default wouldn't be 1 and needs configuring

adam-cattermole (approved):
There's a few things to address but I think we can follow up, nice job getting it here

src/v2/kuadrant/pipeline/tasks/ratelimit.rs
adam-cattermole: The way I thought it worked was this:
- Top-level predicates must all be met to be able to apply the action at all
- For each conditional data item; IF all of it's predicates apply
  - Add the conditional data entry

I thought that's what this did at least here https://github.com/Kuadrant/wasm-shim/blob/main/src/ratelimit_action.rs#L92-L106 (no idea why the old code isn't using `predicates.apply()` there though anyway)

PR #251, opened 2025-10-31
Auth task

adam-cattermole:
The AuthTask is added here; handling storing data through a new StoreDataTask.

### DIFF
--- a/src/v2/kuadrant/pipeline/tasks/auth.rs
+++ b/src/v2/kuadrant/pipeline/tasks/auth.rs
@@ -8,7 +8,7 @@ use crate::v2::kuadrant::pipeline::tasks::{
 use crate::v2::kuadrant::ReqRespCtx;
 use crate::v2::services::{AuthService, Service};
 use chrono::{DateTime, FixedOffset};
-use log::warn;
+use log::{error, warn};
 use prost_types::value::Kind;
 use std::rc::Rc;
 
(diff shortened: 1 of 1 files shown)


src/v2/services/mod.rs
```diff
@@ -38,4 +53,13 @@ pub trait Service {
...
     fn parse_message(&self, message: Vec<u8>) -> Result<Self::Response, ServiceError>;
+
+    fn get_response(
+        &self,
+        ctx: &mut ReqRespCtx,
+        response_size: usize,
+    ) -> Result<Self::Response, ServiceError> {
+        let message = ctx.get_grpc_response(response_size)?;
+        self.parse_message(message)
+    }
```
adam-cattermole: I think this is nice to be symmetrical to the dispatch, where it takes the ctx and handles retrieving the response through that and parsing to `Self::Response` type

src/v2/kuadrant/pipeline/tasks/auth.rs
```diff
@@ -0,0 +1,238 @@
...
+                process_response: Box::new(move |ctx, status_code, response_size| {
+                    if status_code != proxy_wasm::types::Status::Ok as u32 {
+                        // todo(refactor): failure case
+                        return TaskOutcome::Failed;
+                    }
+
+                    match service.get_response(ctx, response_size) {
+                        Ok(parsed) => process_auth_response(parsed),
+                        Err(_e) => TaskOutcome::Failed,
+                    }
```
adam-cattermole: The `status_code` gets passed here so we can decide what to do with it; and then defers to the service to grab the message

Issue #255, opened 2025-11-06
Finalise RateLimitTask


adam-cattermole:
There's some outstanding feedback from the original PR #250 that needs addressing.

I also believe we need to think about how the data is being fetched to follow the new model instead of the old; and ensure it's handling all situations.

PR #261, opened 2025-11-06
Fix RL task descriptor building

didierofrivia:
This PR fixes the wrong logic for building RL descriptors. It builds the descriptors if all of the (action) top level predicates apply and filters the conditional data if its predicates apply. It also leverages the new code for testing predicates.

It partially addresses https://github.com/Kuadrant/wasm-shim/issues/255, it's missing the handling the outstanding parts in processing the response and generating additional follow up tasks.

### DIFF
--- a/src/v2/kuadrant/pipeline/tasks/ratelimit.rs
+++ b/src/v2/kuadrant/pipeline/tasks/ratelimit.rs
@@ -72,9 +72,6 @@ pub struct RateLimitTask {
     // Conditional data for building descriptors
     conditional_data_sets: Vec<ConditionalData>,
     predicates: Vec<Predicate>,
-
-    // Default hits addend
-    default_hits_addend: u32,
 }
 
 /// Creates a new RL task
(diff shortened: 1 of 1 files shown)


alexsnaps:
Maybe @adam-cattermole can review this... I'm confused.

adam-cattermole:
One thing that has changed in the recent merges is there is no longer a `prepare` phase; this is tried up front on task creation - see [`AuthTask`](https://github.com/Kuadrant/wasm-shim/blob/alternate-design/src/v2/kuadrant/pipeline/tasks/auth.rs#L44-L84) (albeit I now realise we should change the name of this function given it has a dependencies field).

When the `AuthTask` is created we try to cache all data; if it's request data it'll then be available on task apply (this is to handle the hypothetical case we want to do auth in the response phase). Because we still have the dependencies check here in [the executor ](https://github.com/Kuadrant/wasm-shim/blob/alternate-design/src/v2/kuadrant/pipeline/tasks/auth.rs#L44-L84) a `RateLimitTask.apply()` won't run until it's previous dependencies are met. I think this is not handled by the code here right?

i.e. RateLimitTask depends on task running in response phase, but it's conditional data or predicates are related to request phase the data won't be cached?

adam-cattermole (approved):
Nice!

PR #262, opened 2025-11-07
Update to use `AuthTask` `new_with_attributes`

adam-cattermole:
- Start using the `AuthTask` `new_with_attributes` and wire this up in the blueprint
- New log order in tests


didierofrivia (approved):
🥥

PR #268, opened 2025-11-10
Cleanups continued

adam-cattermole:
Fixes for:

- The host splitting behaviour is back
- The docker-compose was not working without entrypoint in my environment
- Missed the resuming of the http requests once we're done `Pause`ing; fixes the deadlocking

E2E green!

### DIFF
--- a/src/v2/filter/kuadrant_filter.rs
+++ b/src/v2/filter/kuadrant_filter.rs
@@ -8,6 +8,7 @@ pub struct KuadrantFilter {
     context_id: u32,
     factory: Rc<PipelineFactory>,
     pipeline: Option<Pipeline>,
+    in_response_phase: bool,
 }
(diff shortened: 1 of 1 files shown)


src/v2/filter/kuadrant_filter.rs
```diff
@@ -32,6 +34,21 @@ impl Context for KuadrantFilter {
...
+                } else {
+                    self.resume_http_request()
+                };
+
+                if let Err(e) = result {
+                    error!(
+                        "#{} failed to resume filter processing: {:?}",
+                        self.context_id, e
+                    );
+                }
```
adam-cattermole: We might have to do more than logging only here; I think maybe we should panic/fatal? I _think_ this will deadlock indefinitely

PR #276, opened 2025-11-13
Fix host setting and lookup in cache

adam-cattermole:
There was a couple bugs in the way we store and read from the cache; in particular with the way things are namespaced as well as the escaping. I've tried to make it consistent so that all props are stored at the location that you'd expect in the cache; BUT are escaped and namespaced when we hit the host

This makes it so:
- auth creates unescaped path `['auth', 'identity', 'userid']`
- we store it in the cache as the path `auth.identity.userid`
- we store it in the host, escaped and prefixed as `kuadrant\.auth\.identity\.userid` (envoy then _actually_ stores it under `filter_state.wasm\.{entry}`
- we lookup from the cache as `auth.identity.userid`
- If cache miss, we lookup from the host with escaped prefixed key `filter_state.wasm\.kuadrant\.auth\.identity\.userid`

This is fixing the authenticated rate limiting guide _and_ ensuring stored props are retrieved from the cache correctly and do not go to the host.

### DIFF
--- a/src/kuadrant/context.rs
+++ b/src/kuadrant/context.rs
@@ -137,7 +137,15 @@ impl ReqRespCtx {
     }
 
     fn store_attribute_bytes(&self, path: &Path, value: Vec<u8>) -> Result<(), AttributeError> {
-        self.backend.set_attribute(path, &value)?;
+        // We store in the filter escaped with the kuadrant namespace
+        // but in the cache it remains unchanged
+        const KUADRANT_NAMESPACE: &str = "kuadrant";
+
+        let escaped_path = path.tokens().join("\\.");
+        let host_path = format!("{}\\.{}", KUADRANT_NAMESPACE, escaped_path);
+
+        self.backend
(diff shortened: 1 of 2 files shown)


src/kuadrant/context.rs
```diff
@@ -137,7 +137,15 @@ impl ReqRespCtx {
...
+        // but in the cache it remains unchanged
+        const KUADRANT_NAMESPACE: &str = "kuadrant";
+
+        let escaped_path = path.tokens().join("\\.");
+        let host_path = format!("{}\\.{}", KUADRANT_NAMESPACE, escaped_path);
+
+        self.backend
+            .set_attribute(&host_path.as_str().into(), &value)?;
         self.cache
             .insert(path.clone(), CachedValue::Bytes(Some(value)))
```
adam-cattermole: Cache uses original path; host uses escaped namespaced path

didierofrivia (approved):
We'll need to write some more tests for these use cases

PR #288, opened 2025-11-21
Add OpenTelemetry Tracing Support

adam-cattermole:
## Changes

- **`TracingService`** (`src/services/tracing.rs`): Exports traces via OTLP gRPC to collector endpoints
- **`BufferingSpanProcessor`** (`src/tracing/processor.rs`): span processor with global buffering
- **Trace Propagation** (`src/tracing/propagation.rs`): W3C Trace Context support via `traceparent`/`tracestate` headers
- **`ExportTracesTask`** (`src/kuadrant/pipeline/tasks/export_traces.rs`): Teardown action that exports buffered spans after request completion
- Pipeline instrumentation with spans for executor phases and filter lifecycle
- Trace resource metadata includes service name, version, git hash, and build profile

## Configuration

```yaml
services:
  my-otlp-collector:
    type: tracing
    endpoint: otlp-collector
    failureMode: allow

observability:
  tracing:
    service: my-otlp-collector
```

### DIFF
--- a/src/kuadrant/context.rs
+++ b/src/kuadrant/context.rs
@@ -49,7 +49,7 @@ impl ReqRespCtx {
         self
     }
 
-    pub fn init_tracing(&mut self) {
+    pub fn extract_trace_context(&mut self) {
         let request_headers: Result<AttributeState<Option<Headers>>, _> =
             self.get_attribute("request.headers");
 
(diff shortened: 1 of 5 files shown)


e2e/basic/docker-compose.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 services:
   envoy:
-    image: envoyproxy/envoy:v1.31-latest
```
adam-cattermole: This needs to be `v1.32+` for the updated wasmtime, which means istio `v1.24+` (`sched_yield` missing in v1.31)

alexsnaps (dismissed):
wcgw?

PR #303, opened 2026-01-07
Store body values in `ReqRespCtx` instead of `Expression`

adam-cattermole:
An `Expression` contains `response_props`, but an `Expression` would live for the lifetime of the filter (post on_configure) but the ``response_props` are tied to the request response lifecycle. We had made some patches to fix the behaviour, but ultimately the original approach didn't make a lot of sense. I have:

- Made it so that the expressions hold only the keys of the body values that it cares about - this still happens once at expression parse time but the lifetime is for the rest of the filter
- The `Action`'s (compiled from config with parsed expressions) extract the set of body values that they are interested in
- The `TokenUsageTask` then stores the values in the `ReqRespCtx` - they will only have the lifetime of the request
- The `Expression` reads these values from the `ReqRespCtx` in `eval`

### DIFF
--- a/src/data/cel.rs
+++ b/src/data/cel.rs
@@ -1042,6 +1042,43 @@ mod tests {
     }
 
     #[test]
+    fn response_body_json_returns_pending_when_not_set() {
+        let expr = Expression::new("responseBodyJSON('bar') == 42").unwrap();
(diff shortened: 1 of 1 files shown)


adam-cattermole:
Rebased on #304

Issue #231, opened 2025-10-09
Added headers should replace existing values, not append


guicassolato:
**Describe the bug**
As stated in the docs for ext_authz[^1], the default behaviour when adding headers to the request – that are not inverted by specifying `ok_response.headers.append: true` in the `CheckResponse` – should be _replace_ existing values. Currently, the wasm-shim _adds_ another header indistinctly[^2], effectively causing the value to be _appended_ to existing ones, comma-separated. This makes the behaviour of the wasm-shim's implementation of ext_authz inconsistent compared to Envoy's[^3].

There's currently an [issue](https://github.com/envoyproxy/envoy/issues/22845) opened in the Envoy repo postulating the default behaviour implemented for ext_authz as a bug, which ended up resulting in docs updates and the new `append_action` field introduced. However, `append_action` has never been implemented for `ok_response.headers`, only for `ok_response. response_headers_to_add`. See more at https://github.com/Kuadrant/authorino/issues/562#issuecomment-3385198656.

[^1]: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-msg-config-core-v3-headervalueoption
[^2]: https://github.com/Kuadrant/wasm-shim/blob/main/src/auth_action.rs → https://github.com/Kuadrant/wasm-shim/blob/f417fc845c6eaa25b66aee45ec330f70be4137dd/src/filter/kuadrant_filter.rs#L514
[^3]: https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/common/ext_authz/ext_authz_grpc_impl.cc#L19-L67

**To Reproduce**
Using Kuadrant,
1. Create an AuthPolicy that sets a success response header `x-header: new value`
2. Send a request in the scope of the AuthPolicy containing `x-header: old value`
3. Check the request that hits the upstream, the `x-header` header will be `old value,new value`

**Expected behavior**
Following the steps to reproduce above, `x-header` should hit the upstream with value set to `new value`

PR #315, opened 2026-03-06
Replace headers default

adam-cattermole:
The changes from #231, which took place in the original code simultaneously to the refactor were not correctly ported to the refactor and omitted in v0.12.0.


KevFan (approved):
Looks good to me 👍

PR #332, opened 2026-04-27
feat: add gRPC well-known attributes (request.grpc)

### LINKED ISSUE #317: [GRPCRoute] gRPC Well-Known Attributes
## Task 10: gRPC Well-Known Attributes

**Parent epic:** Kuadrant/architecture#156  
**RFC:** [0017-grpcroute-support.md](https://github.com/Kuadrant/architecture/blob/main/rfcs/0017-grpcroute-support.md)

## What

Implement `request.grpc.service` and `request.grpc.method` well-known attributes in the WASM shim to improve UX when writing policy conditions and rate limit counters for gRPC traffic. Update RFC 0002 specification to document the new attributes.

## Why

While GRPCRoute support works using `request.url_path` patterns (e.g., `request.url_path == '/UserService/GetUser'`), policy authors benefit from dedicated attributes for better readability and easier policy authoring:

```yaml
# Instead of:
when:
  - predicate: request.url_path == '/UserService/GetUser'

# Policy authors can write:
when:
  - predicate: has(request.grpc) && request.grpc.service == 'UserService'
  - predicate: has(request.grpc) && request.grpc.method == 'GetUser'
```

### wasm-shim Implementation

- [ ] Implement gRPC request detection via `content-type: application/grpc` header
- [ ] Implement path parsing for `/Service/Method` format
- [ ] Inject `grpc` key into `request` CEL variable as a map with `service` and `method` fields
- [ ] For non-gRPC requests OR malformed paths, `grpc` key is absent from `request` map
- [ ] Unit tests for path parsing (exact service+method, service-only, method-only, malformed)
- [ ] Unit tests for gRPC detection via content-type header
- [ ] Unit tests verifying `has(request.grpc)` returns false for HTTP requests
- [ ] Unit tests verifying accessing `request.grpc` without `has()` check errors on HTTP traffic
- [ ] Integration tests with actual gRPC traffic

### RFC 0002 Updates

- [ ] Add "gRPC Well-Known Attributes" section to RFC 0002
- [ ] Document `request.grpc.service` attribute with type, description, examples
- [ ] Document `request.grpc.method` attribute with type, description, examples
- [ ] Document presence checking via `has(request.grpc)` macro
- [ ] Document behavior with non-gRPC requests (key is absent)
- [ ] Document best practices for mixed HTTP/gRPC traffic
- [ ] Add implementation notes referencing WASM shim

## Notes

**This task can be developed in parallel with operator work (Tasks 3-7).** The WASM shim implementation is independent, and the kuadrant-operator continues using `request.url_path` patterns internally.

**Key files (wasm-shim):**
- `src/data/attribute.rs` - Well-known attributes implementation
- `src/filter/http_context.rs` - HTTP filter context with header access
- `src/data/cel.rs` - CEL expression evaluation

**Implementation approach:**
```rust
// Detect gRPC requests
fn is_grpc_request(&self) -> bool {
    self.get_http_request_header("content-type")
        .map(|ct| ct.starts_with("application/grpc"))
        .unwrap_or(false)
}

// Parse path to extract service and method
fn extract_grpc_service_method(path: &str) -> Option<(String, String)> {
    let parts: Vec<&str> = path.trim_start_matches('/').split('/').collect();
    if parts.len() == 2 {
        Some((parts[0].to_string(), parts[1].to_string()))
    } else {
        None
    }
}
```

**CEL Map-Based Approach:**

The `request.grpc` attribute uses standard CEL map semantics for presence checking:
- For **gRPC requests**: `has(request.grpc)` returns `true`, access fields via `request.grpc.service`
- For **HTTP requests** OR **malformed gRPC paths**: `has(request.grpc)` returns `false`, the key is absent

Policy authors must use `has()` to check presence before accessing fields:
```yaml
# Correct syntax (safe for mixed HTTP/gRPC traffic):
when:
  - predicate: has(request.grpc) && request.grpc.service == 'UserService'
  - predicate: has(request.grpc) && request.grpc.method == 'GetUser'

# Incorrect (will error on HTTP requests):
when:
  - predicate: request.grpc.service == 'UserService'  # ERROR: key not found for HTTP
```

**Critical:** These attributes are **optional** for policy authors. All existing predicates using `request.url_path` continue to work. The kuadrant-operator internal predicate generation continues using `request.url_path` to avoid hard dependency on this task.

**No changes needed in kuadrant-operator:** The operator continues generating predicates using `request.url_path` patterns. These well-known attributes are UX conveniences for policy authors writing custom `when` conditions and rate limit counters, not for internal operator use.

**Edge case handling:**
- For non-gRPC requests OR malformed paths, the `grpc` key is absent from the `request` map
- Always use `has(request.grpc)` check before accessing `request.grpc` to avoid errors on HTTP traffic
- For route-level policies targeting GRPCRoute directly where all traffic is guaranteed to be gRPC, the `has()` check may be omitted, but including it is safer
- Avoid negation patterns like `request.grpc.service != 'AdminService'` without a `has()` guard as these will error on HTTP traffic

**Testing coverage:**
- Path parsing: `/Service/Method`, `/Package.Service/Method`, malformed paths
- Content-type detection: `application/grpc`, `application/grpc+proto`, `application/json`
- HTTP vs gRPC: `has(request.grpc)` should return `false` for HTTP requests
- CEL evaluation: Verify that accessing `request.grpc` without `has()` check errors appropriately on HTTP traffic
- Integration: Real gRPC traffic using grpcbin backend (from Task 2)

mikenairn:
## Summary

- Add `request.grpc.service` and `request.grpc.method` as computed CEL attributes
- gRPC requests detected via `content-type: application/grpc` header and `/Service/Method` path parsing
- For non-gRPC requests, `request.grpc` is absent — `has(request.grpc)` returns `false`
- New `src/data/grpc.rs` module with detection and parsing functions
- No CEL optional types extension required — uses standard `has()` macro for presence checking

### Usage

```yaml
# Check if request is gRPC
when:
  - predicate: has(request.grpc) && request.grpc.service == 'UserService'
  - predicate: has(request.grpc) && request.grpc.method == 'GetUser'
```

## Test plan

- [x] Unit tests for `is_grpc_content_type()` (6 cases)
- [x] Unit tests for `parse_grpc_path()` (8 cases)
- [x] CEL integration: `request.grpc.service` matches on gRPC request
- [x] CEL integration: `request.grpc.method` matches on gRPC request
- [x] CEL integration: `has(request.grpc)` returns false for HTTP requests
- [x] CEL integration: `has(request.grpc)` returns false for malformed gRPC path
- [x] CEL integration: `has(request.grpc) && request.grpc.service == 'X'` guard pattern works
- [x] CEL integration: guard pattern returns false on HTTP traffic

Resolves https://github.com/Kuadrant/wasm-shim/issues/317

## Summary by CodeRabbit

* **New Features**
  * CEL policies are now gRPC-aware, exposing request gRPC service/method for more precise policy matching.
  * Added gRPC detection and path-parsing utilities.

* **Chores**
  * Enhanced request header and attribute-map retrieval helpers across host/resolver implementations to support gRPC logic.

* **Tests**
  * Added tests for gRPC content-type detection, path parsing, attribute filtering and correct gRPC match/guard behaviour.

### DIFF
--- a/src/data/cel.rs
+++ b/src/data/cel.rs
@@ -158,14 +158,12 @@ impl Expression {
             &mut Vec::default(),
         );
 
-        let needs_grpc = props
-            .iter()
-            .any(|tokens| tokens.len() >= 2 && tokens[0] == "request" && tokens[1] == "grpc");
-
+        let mut needs_grpc = false;
         let mut attributes: Vec<Attribute> = props
             .into_iter()
             .filter_map(|tokens| {
                 if tokens.len() >= 2 && tokens[0] == "request" && tokens[1] == "grpc" {
+                    needs_grpc = true;
                     return None;
                 }
                 let root = tokens.first()?;
@@ -276,27 +274,23 @@ impl Expression {
     }
 
     fn resolve_grpc(&self, req_ctx: &ReqRespCtx) -> Option<Value> {
-        let content_type: Option<String> = req_ctx
-            .get_attribute::<Headers>("request.headers")
-            .ok()
-            .and_then(|state| match state {
-                AttributeState::Available(Some(headers)) => {
-                    headers.get("content-type").map(|s| s.to_string())
-                }
-                _ => None,
-            });
+        let content_type = req_ctx.get_request_header("content-type");
 
         if !content_type.as_deref().is_some_and(is_grpc_content_type) {
             return None;
         }
 
-        let url_path: Option<String> = req_ctx
-            .get_attribute::<String>("request.url_path")
-            .ok()
-            .and_then(|state| match state {
-                AttributeState::Available(opt) => opt,
-                _ => None,
-            });
+        let url_path: Option<String> = match req_ctx.get_attribute::<String>("request.url_path") {
+            Ok(AttributeState::Available(opt)) => opt,
+            Ok(AttributeState::Pending) => {
+                warn!("request.url_path is pending during gRPC attribute resolution");
+                None
+            }
+            Err(e) => {
+                warn!("failed to get request.url_path for gRPC resolution: {e}");
+                None
+            }
+        };
 
         let (service, method) = match url_path.as_deref().and_then(parse_grpc_path) {
             Some((s, m)) => (Value::String(s.into()), Value::String(m.into())),
@@ -1425,6 +1419,23 @@ mod tests {
             );
         let ctx = ReqRespCtx::new(Arc::new(mock_host));
         let predicate = Predicate::new("request.grpc.method == 'GetUser'").expect("valid CEL");
+        assert_eq!(
+            predicate.test(&ctx).expect("must evaluate"),
+            AttributeState::Available(true)
+        );
+    }
+
+    #[test]
+    fn grpc_has_true_for_grpc_request() {
+        let headers = vec![("content-type".to_string(), "application/grpc".to_string())];
+        let mock_host = MockWasmHost::new()
+            .with_map("request.headers".to_string(), headers)
+            .with_property(
+                "request.url_path".into(),
+                "/UserService/GetUser".bytes().collect(),
+            );
+        let ctx = ReqRespCtx::new(Arc::new(mock_host));
+        let predicate = Predicate::new("has(request.grpc)").expect("valid CEL");
         assert_eq!(
             predicate.test(&ctx).expect("must evaluate"),
             AttributeState::Available(true)
--- a/src/data/cel.rs
+++ b/src/data/cel.rs
@@ -1,5 +1,6 @@
 use crate::data::attribute::{AttributeError, AttributeState, Path};
 use crate::data::cel::errors::{CelError, EvaluationError};
+use crate::data::grpc::{is_grpc_content_type, parse_grpc_path};
 use crate::data::Headers;
 use crate::kuadrant::ReqRespCtx;
 use cel::common::ast::{EntryExpr, Expr, IdedExpr, LiteralValue};
@@ -104,6 +105,7 @@ pub struct Expression {
     body_values: Vec<String>,
     expression: CelExpression,
     extended: bool,
+    needs_grpc: bool,
 }
 
 pub type EvalResult = Result<AttributeState<Value>, CelError>;
@@ -112,8 +114,8 @@ impl Display for Expression {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         write!(
             f,
-            "Expression {{ expression: {:?}, attributes: {:?}, extended: {} }}",
-            self.expression, self.attributes, self.extended
+            "Expression {{ expression: {:?}, attributes: {:?}, extended: {}, needs_grpc: {} }}",
+            self.expression, self.attributes, self.extended, self.needs_grpc
         )
     }
 }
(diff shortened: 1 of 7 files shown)


src/data/cel.rs
```diff
@@ -249,6 +274,41 @@ impl Expression {
...
+    fn resolve_grpc(&self, req_ctx: &ReqRespCtx) -> Option<Value> {
+        let content_type: Option<String> = req_ctx
+            .get_attribute::<Headers>("request.headers")
+            .ok()
+            .and_then(|state| match state {
+                AttributeState::Available(Some(headers)) => {
+                    headers.get("content-type").map(|s| s.to_string())
+                }
+                _ => None,
+            });
```
alexsnaps: Avoid loading all headers here, this is potentially really costly. Instead only try getting to the `content-type` directly as this is on the hot path

src/data/cel.rs
```diff
@@ -155,9 +158,16 @@ impl Expression {
...
+        let needs_grpc = props
+            .iter()
+            .any(|tokens| tokens.len() >= 2 && tokens[0] == "request" && tokens[1] == "grpc");
+
         let mut attributes: Vec<Attribute> = props
             .into_iter()
             .filter_map(|tokens| {
+                if tokens.len() >= 2 && tokens[0] == "request" && tokens[1] == "grpc" {
+                    return None;
+                }
```
alexsnaps: We should avoid iterating twice over the props here, there isn't really any reason to make this `O(2n)` is there? Not on the hot path, so not a _huge_ deal, but...

src/data/cel.rs
```diff
@@ -177,6 +187,7 @@ impl Expression {
             body_values: response_props,
             expression,
             extended,
+            needs_grpc,
```
alexsnaps, line 188: I don't think we actually _need_ this. We probably would need to adapt the logic around e.g. [`auth`](https://github.com/Kuadrant/wasm-shim/blob/main/src/kuadrant/context.rs#L187-L194) to being able to deal with this, but we _also_ need the `has` check to succeed, so that the path to these properties can't be entirely lazy. 

So, let's keep this as is for now, and we'll revisit when `request` becomes a proper gRPC `Message` and isn't this `Map` anymore.

alexsnaps (changes requested):

src/data/cel.rs
mikenairn: Added AttributeResolver::get_attribute_map_value() and ReqRespCtx::get_request_header() to fetch a single header directly from Envoy rather than loading the entire headers map

src/data/cel.rs
mikenairn: updated to remove the extra iteration of props

src/kuadrant/resolver/mock.rs
```diff
@@ -109,6 +109,15 @@ impl AttributeResolver for MockWasmHost {
...
     }
 
+    fn get_attribute_map_value(
+        &self,
+        map_type: proxy_wasm::types::MapType,
+        key: &str,
+    ) -> Result<Option<String>, AttributeError> {
+        let map = self.get_attribute_map(map_type)?;
+        Ok(map.into_iter().find(|(k, _)| k == key).map(|(_, v)| v))
+    }
```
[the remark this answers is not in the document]
alexsnaps: Envoy guarantees theses are HTTP/2 headers, i.e. lower-cased.

alexsnaps (approved):

PR #348, opened 2026-05-19
ci: trigger test workflow on release branch pushes

thomasmaas:
## Summary

Adds `release-*` to the push branch trigger in `tests.yaml` so CI runs when commits land on release branches. The pull_request trigger already uses a wildcard.

Part of the RFC 0018 migration (Kuadrant/kuadrant-operator#1981).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -3,7 +3,7 @@ name: Tests
 
 on:
   push:
-    branches: ["main"]
+    branches: ["main", "release-*"]
   pull_request:
     branches: ["*"]
   merge_group:


alexsnaps (approved):

.github/workflows/tests.yaml
```diff
@@ -3,7 +3,7 @@ name: Tests
 
 on:
   push:
-    branches: ["main"]
+    branches: ["main", "release-*"]
```
alexsnaps, line 6: just to be sure I understand, this is for _branches_ yes? Not tags? I guess they will be on `main`? Tho, given this is a rust code base, it might be a tag to a commit not on any branches, right?

thomasmaas: this is for branches, correct.

PR #359, opened 2026-05-27
Refactor `SendReplyTask`

### LINKED ISSUE #364: Make Predicates universal and promote actions to top-level
- Make predicates universal across action tasks (SendReply/Headers/StoreData) - right now these are evaluated to build the task - instead the matches could map (maybe try_from Operation -> Task?) to the task type and try apply/requeue to seriously simplify that function
- Promote Deny/Headers/Store/Fail to top-level actions and restructure the Action struct to handle any of the types - right now we support one use-case as top-level which is grpc. For extensions to be able to register a single 'add header action' etc we can update the model

alexsnaps:
Not at all as far as where I was hoping to get, but that's the idea. 
When that's done for all other "`Action`", `DynamicTask` won't need the `Action`s anymore, but the `Pipeline` could be constructed with "only `Task`s". And all the `impl TryFrom<Value> for` can be :fire: too.

@adam-cattermole Do not expand on this, but have a look at the `Store` for `requestBodyJSON('/model')` as discussed. I'll keep moving this PR forward (my) tomorrow. Now all input is welcome already tho!

## Summary by CodeRabbit

* **Refactor**
  * Refactored denial response handling to improve evaluation flexibility.
  * Adjusted denial metrics recording to occur at execution time for enhanced accuracy.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/wasm-shim/pull/359?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/services/mod.rs
+++ b/src/services/mod.rs
@@ -6,7 +6,9 @@ use std::{rc::Rc, time::Duration};
 mod dynamic;
 mod tracing;
 
-pub use dynamic::converters::{cel_value_to_header_pairs, MessageConverter};
+pub use dynamic::converters::{
+    cel_value_to_header_pairs, deny_response_struct_def, MessageConverter,
+};
 pub use dynamic::DynamicService;
 pub use tracing::TracingService;
 
--- a/src/kuadrant/pipeline/tasks/send_reply.rs
+++ b/src/kuadrant/pipeline/tasks/send_reply.rs
@@ -29,7 +29,6 @@ pub struct SendReplyTask {
 
 impl SendReplyTask {
     pub fn new(status_code: u32, headers: Vec<(String, String)>, body: Option<String>) -> Self {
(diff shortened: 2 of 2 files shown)


src/services/mod.rs
```diff
@@ -6,7 +6,9 @@ use std::{rc::Rc, time::Duration};
 mod dynamic;
 mod tracing;
 
-pub use dynamic::converters::{cel_value_to_header_pairs, MessageConverter};
+pub use dynamic::converters::{
+    cel_value_to_header_pairs, deny_response_struct_def, MessageConverter,
```
alexsnaps, line 10: Slightly confused as to why `deny_response_struct_def` is declared in that module...?

adam-cattermole (approved):

PR #367, opened 2026-05-29
Add linker arg for rust 1.96

adam-cattermole:
I think the correct fix is to update in proxy-wasm-rust-sdk and add specific link attributes https://blog.rust-lang.org/2026/04/04/changes-to-webassembly-targets-and-handling-undefined-symbols/


Boomatang (approved):
All check are passing, happy to lgtm

PR #369, opened 2026-05-29
Remove implicit sequential dependency from typed action

adam-cattermole:
DynamicTasks are treated as sequential implicitly rather than trusting the configuration to explicitly define dependence. Their onreply tasks also are tried to execute on response.. but then will enforce sequential ordering if requeued.

Once we migrate from legacy config we can remove the dependencies field entirely and have it explicitly defined.

## Release Notes

* **Bug Fixes**
  * Improved action dependency handling in the pipeline compilation process for better execution flow and consistency between different action types.

### DIFF
--- a/src/kuadrant/pipeline/blueprint.rs
+++ b/src/kuadrant/pipeline/blueprint.rs
@@ -186,13 +186,13 @@ impl Blueprint {
             .enumerate()
             .map(|(i, action_config)| {
                 let id = i.to_string();
-                let dependencies = if i > 0 {
-                    vec![(i - 1).to_string()]
-                } else {
-                    vec![]
-                };
                 match action_config {
                     configuration::ActionConfig::Legacy(action) => {
+                        let dependencies = if i > 0 {
+                            vec![(i - 1).to_string()]
+                        } else {
+                            vec![]
+                        };
                         let legacy_request_data: Vec<((String, String), String)> = request_data
                             .iter()
                             .map(|(key, expr)| (key.clone(), expr.source().to_string()))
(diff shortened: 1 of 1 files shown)


src/kuadrant/pipeline/blueprint.rs
```diff
@@ -186,21 +186,21 @@ impl Blueprint {
...
-                } else {
-                    vec![]
-                };
                 match action_config {
                     configuration::ActionConfig::Legacy(action) => {
+                        let dependencies = if i > 0 {
+                            vec![(i - 1).to_string()]
+                        } else {
+                            vec![]
+                        };
```
adam-cattermole, line 195: This keeps the dependencies between legacy types but not for the new types.. but `dependencies` should disappear once we burn `Legacy`

didierofrivia: Burning `Legacy` involves changing the way Kuadrant Operator builds the config?

adam-cattermole: Yep exactly, working on that at the minute

didierofrivia: what I mean is, the change is required until we burn `Legacy` that in part needs to change the way the Operator builds the config (?)

adam-cattermole: Oh, sorry yep I follow. Right now this change only affects extensions that go through the non-legacy route. The dependence between actions has to be encoded as part of the actual actions themselves (either through nested onReply, _or_ through dependence through exporting data etc that the other task requires)

Right now anyone who configures tasks using an extension has their tasks dependent on the previous one completing, even though there's no inherent dependency - which doesn't make a whole lot of sense.

In the kuadrant-operator we'll have to encode a predicate on `auth.` presence in ratelimit actions to enforce this sequential behaviour, _but_ if an extension author wants to fire off 3 gRPC requests in a row without waiting sequentially for each of their responses, it'll be possible after this change.

didierofrivia (approved):

PR #379, opened 2026-06-22
Reduce tracing log occurrence

adam-cattermole:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved span buffer full-state warning to emit only once per full period rather than continuously, reducing warning spam whilst maintaining maximum buffer capacity and dropping oldest spans as needed.

### DIFF
--- a/src/tracing/processor.rs
+++ b/src/tracing/processor.rs
@@ -37,8 +37,14 @@ impl SpanProcessor for SpanProcessorHandle {
 }
 
 #[derive(Debug)]
+struct BufferState {
+    queue: VecDeque<SpanData>,
+    logged_full: bool,
+}
+
+#[derive(Debug)]
 pub struct BufferingSpanProcessor {
-    buffer: Mutex<VecDeque<SpanData>>,
+    buffer: Mutex<BufferState>,
     max_buffer_size: usize,
 }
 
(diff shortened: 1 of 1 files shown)


src/tracing/processor.rs
```diff
@@ -50,37 +56,38 @@ impl BufferingSpanProcessor {
...
     }
 
     /// Take all pending spans from the buffer, clearing it
     pub fn take_pending_spans(&self) -> Vec<SpanData> {
-        let mut buffer = self.buffer.lock().unwrap_or_else(|e| e.into_inner());
-        std::mem::take(&mut *buffer).into()
+        let mut state = self.buffer.lock().unwrap_or_else(|e| e.into_inner());
+        state.logged_full = false;
+        std::mem::take(&mut state.queue).into()
     }
```
adam-cattermole, line 72: Here and `clear` are the two locations where we remove spans so instead of handling bouncing around the boundary by 5/10 spans instead we reset on each 'flush'.

src/tracing/processor.rs
```diff
@@ -101,18 +108,21 @@ impl SpanProcessor for BufferingSpanProcessor {
...
 
         // FIFO
-        if buffer.len() >= self.max_buffer_size {
-            tracing::warn!(
-                "Span buffer full ({}), dropping oldest span",
-                self.max_buffer_size
-            );
-            buffer.pop_front();
+        if state.queue.len() >= self.max_buffer_size {
+            if !state.logged_full {
```
alexsnaps, line 115: If clippy is happy with this nested `if`s, then so am I  ¯\_(ツ)_/¯

alexsnaps (approved):
Real nitpick: 

 - `impl Default for BufferState` (given the cap is hard coded anyways)
 - Replace held value with default on `clear`, and you got rid of the `logged_full` "manual" reset
 - You _could_ even do a `mem::replace` with `Default::default()` on the value within the mutex, so you don't have to think about the `bool` (other than on logging)

adam-cattermole:
> Replace held value with default on clear, and you got rid of the logged_full "manual" reset

Hmm won't this mean we allocate a new `VecDequeue` and drop the old one vs `clear`ing the original? I don't know which is more performant or if they end up the same

alexsnaps:
>Hmm won't this mean we allocate a new `VecDequeue` and drop the old one vs `clear`ing the original? I don't know which is more performant or if they end up the same

Yes, tho we'd probably need to read the resulting assembly to make sure... 
Another mean tho, is to encapsulate all that behavior _in_ `BufferState` over "simply" treating it as a "tuple". But, again, this is me being real nitpicky - I realize that. My goal here was mostly in "the benefit of primarily removing the error prone flipping of `bool`". I honestly don't think saving the overhead of the alloc is worth it in this context (i.e. once we flush the buffers, once per request _and have to wait to send them over the wires and get a response_. 

Really, feel free to ignore tho.

adam-cattermole:
Gave it a go but ended up with a strange situation trying to expose the preset max capacity of the queue and it became a bit convoluted - was nice from the perspective of not managing the bool but yeah the capacity wasn't nicely encapsulated anymore 🤷 so I'll just merge as is

PR #380, opened 2026-06-25
Update incorrect workflow token

adam-cattermole:
Relates to #365 - the `WORKFLOWS_PAT` doesn't have access to this repo

## Summary by CodeRabbit

* **Chores**
  * Improved the automated release process to handle release branch publishing more reliably.
  * Updated the generated release title formatting for consistency.

### DIFF
--- a/.github/workflows/automated-release.yaml
+++ b/.github/workflows/automated-release.yaml
@@ -23,13 +23,13 @@ jobs:
         uses: actions/checkout@v6
         with:
           ref: ${{ github.event.inputs.gitRef }}
-          token: ${{ secrets.KUADRANT_WORKFLOWS_PAT }}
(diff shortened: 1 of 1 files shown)


didierofrivia (approved):
I'm approving this PR, however alongside with https://github.com/Kuadrant/wasm-shim/pull/365/ ,  is screaming at us that we are not managing well our PATs since CNCF conditioned the rotation to 365 days tops.

adam-cattermole:
> I'm approving this PR, however alongside with https://github.com/Kuadrant/wasm-shim/pull/365 , is screaming at us that we are not managing well our PATs since CNCF conditioned the rotation to 365 days tops.

Yep agreed - I think these ones do have a 365 day max period before they expire - but we should likely look into slightly finer-grained options here and splitting into several tokens in my opinion

PR #385, opened 2026-06-25
[Release] WASM Shim v0.14.1

adam-cattermole:


thomasmaas:
Wondering if [this change](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/305) will bite us

adam-cattermole:
> Wondering if https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/305 will bite us

Could you elaborate @thomasmaas?

thomasmaas:
If there is any logic in authorino or limitador that depends on not differentiating between between missing and empty headers?

adam-cattermole:
> If there is any logic in authorino or limitador that depends on not differentiating between between missing and empty headers?

I'm not sure I'm following. As far as I can tell the only use of the `get_map_value` is from here, and we correctly validate `None` as not present/not a gRPC content-type:

https://github.com/Kuadrant/wasm-shim/blob/7d59d53c83075d977f463ead32a792f295e85b10/src/data/cel.rs#L333-L337

For our other header access we leverage `get_map` which already allows/supports headers with empty values

didierofrivia (approved):

PR #386, opened 2026-07-08
Context stored values support pending through RAII path reservations

adam-cattermole:
## Summary by CodeRabbit

* **New Features**
  * Improved handling of pending values so reserved paths are tracked cleanly and stored values can be resolved more reliably.
  * Request- and response-body dependencies are now considered when evaluating conditions, allowing tasks to wait for the right data before proceeding.

* **Bug Fixes**
  * Fixed stored-value lookups so existing values take priority over reservations and sibling values are no longer blocked.
  * Requeue and failure behaviour is now more accurate when body data is still streaming or has finished.

### DIFF
--- a/src/data/cel.rs
+++ b/src/data/cel.rs
@@ -257,12 +257,12 @@ impl Expression {
             }
         }
 
-        for path in req_ctx.stored_value_paths() {
+        for path in req_ctx.values.paths() {
             let root = path.split('.').next().unwrap_or(path);
             if root == path {
                 let key: Key = root.into();
                 if !map.contains_key(&key) {
-                    if let Some(value) = req_ctx.get_stored_value(path) {
+                    if let Some(value) = req_ctx.values.get(path) {
                         cel_ctx.add_variable_from_value(root, value.clone());
                     }
                 }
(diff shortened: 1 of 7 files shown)


src/data/cel.rs
```diff
@@ -558,6 +566,14 @@ impl Predicate {
...
         self.expression.response_body_values()
     }
+
+    pub fn has_request_body_deps(&self) -> bool {
+        self.expression.has_request_body_deps()
+    }
+
+    pub fn has_response_body_deps(&self) -> bool {
+        self.expression.has_response_body_deps()
+    }
```
adam-cattermole, line 576: Until we have both SSE and regular body path extraction going through the StoreTask, we need a way to know that no more data is coming. Once SSE is ported to the StoreTask the reservation system will capture this so we can drop these checks (and the end_of_stream check) I think so that it's all handled by the reservation

src/data/cel.rs
```diff
@@ -1862,12 +1881,73 @@ mod tests {
...
+    fn dropped_reservation_is_not_pending() {
+        let req_ctx = ReqRespCtx::new(Arc::new(MockWasmHost::new()));
+        let mut cel_ctx = cel::Context::default();
+
+        let reservation = req_ctx.values.reserve("myvar.value".to_string());
+        drop(reservation);
+
+        let predicate = Predicate::new("myvar.value == null").expect("valid CEL");
+        let result = predicate.test(&req_ctx, &mut cel_ctx);
+        assert!(result.is_err());
```
adam-cattermole, line 1934: The expression fails with `UndeclaredValue(myvar)` as it's no longer reserved/pending and won't be filled, so any dependent tasks CEL expressions should rightfully fail to evaluate

alexsnaps (approved):

PR #397, opened 2026-07-24
Case-insensitive hostname lookup

adam-cattermole:
## Summary by CodeRabbit

- **Bug Fixes**
  - Hostname matching is now case-insensitive, providing consistent routing for uppercase and mixed-case hostnames.
  - Wildcard hostname matching now works regardless of hostname letter casing.
  - Hostnames containing ports are normalised correctly before matching.


adam-cattermole:
Pfft auto-merged when I switched the base branch other than main, I'll reopen

https://github.com/Kuadrant/wasm-shim/pull/405

PR #404, opened 2026-07-29
Update dockerfile

adam-cattermole:
Missed the Dockerfile builds in #399 and the image build failures weren't detected by CI as it wasn't originating from the org

I've enabled in CI a build (but not push) for PRs

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -3,8 +3,10 @@ name: Build WASM Image
 
 on:
   push:
-    branches: ['*']
-    tags: ['*']
+    branches: ["*"]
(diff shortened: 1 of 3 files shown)


.github/workflows/build-image.yaml
```diff
@@ -16,6 +18,7 @@ env:
 jobs:
   build:
     name: Build
+    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.repository }}
```
adam-cattermole, line 21: Skip when it's coming from this repo (we already have build on push)

alexsnaps (approved):

PR #407, opened 2026-07-29
Teardown no longer holds local tasks

### LINKED ISSUE #394: Path reservations cause infinite requeue loop for failed requests (429/401), dropping tracing spans
## Summary

  Since [#386](https://github.com/Kuadrant/wasm-shim/pull/386), the `kuadrant_filter` tracing span is missing for requests that result in 429 (rate limited) or 401 (auth denied) responses. Successful 200 requests produce the expected span. This causes data plane tracing tests to fail because traces cannot be found by `request_id` tag for failed requests.

  ## Root Cause
  
  PR #386 introduced RAII path reservations in `ValueStore`. When a `StoreTask` is created, it reserves its destination path via `ctx.values.reserve(path)`. Any CEL expression referencing that reserved-but-not-yet-stored path now returns `AttributeState::Pending`.

  The same PR changed the requeue logic in `DynamicTask`, `FailTask`, and `StoreTask` to only transition from `Pending` to `Failed` when body-stream dependencies are exhausted:

  ```rust
  Ok(AttributeState::Pending) => {
      if (self.predicate.has_request_body_deps() && ctx.request_body.is_end_of_stream())
          || (self.predicate.has_response_body_deps() && ctx.response_body.is_end_of_stream())
      {
          TaskOutcome::Failed
      } else {
          TaskOutcome::Requeued(vec![self])
      }
  }
  ```
For tasks whose predicates reference a reserved path (not a body field), `both has_request_body_deps()` and `has_response_body_deps()` return `false`. The fail condition is never satisfied, so the task is always requeued.

On the success path (200), this works fine — the `StoreTask` eventually stores the value, clears the pending state, and dependent tasks proceed. On the failure path (429/401), the pipeline short-circuits before the store happens. The `PathReservation` is still held by the unconsumed `StoreTask`, so dependent tasks see `Pending` indefinitely and loop forever. This prevents the pipeline from completing and the `kuadrant_filter` span from being reported to the tracing backend.

  Steps to Reproduce

  1. Deploy Kuadrant with tracing enabled (Jaeger/Tempo)
  2. Configure an AuthPolicy and RateLimitPolicy on a route
  3. Send a request that triggers a 429 (exhaust rate limit) or 401 (no auth credentials)
  4. Query the tracing backend for `service=wasm-shim` with the `request_id` from the response
  5. No trace is found — the `kuadrant_filter` span is missing

  Expected Behavior

The `kuadrant_filter` span should be present for all requests regardless of response status code. 

  Actual Behavior

  - 200 responses: `kuadrant_filter` span present with full span hierarchy (correct)
  - 429 responses: no trace found for the `request_id` (regression)
  - 401 responses: no trace found for the `request_id` (regression)

adam-cattermole:
When we migrated `onReply` actions to become fully fledged tasks the reference to their spans are held in the local var `tasks_to_process`. This prevents these spans from being dropped and closed _prior_ to calling the `execute_teardown` which sends the traces causing the overarching `kuadrant_filter` span to be held open.

I've moved the processing to it's own function which is unchanged, but ensures the local ref to the current eval loop is no longer being held (in a different scope) when we process the teardown tasks

### DIFF
--- a/crates/kuadrant-filter/src/kuadrant/pipeline/executor.rs
+++ b/crates/kuadrant-filter/src/kuadrant/pipeline/executor.rs
@@ -14,6 +14,11 @@ pub enum PipelineState {
     Completed { should_resume: bool },
 }
 
+enum ProcessOutcome {
+    Continue,
+    Terminated,
+}
+
 pub struct Pipeline {
     pub ctx: ReqRespCtx,
     task_queue: Vec<Box<dyn Task>>,
@@ -100,10 +105,8 @@ impl Pipeline {
         }
     }
 
-    pub fn eval(mut self) -> PipelineState {
-        let tasks_to_process: Vec<_> = self.task_queue.drain(..).collect();
-
-        for task in tasks_to_process {
+    fn process_tasks(&mut self, tasks: Vec<Box<dyn Task>>) -> ProcessOutcome {
+        for task in tasks {
             if task
                 .dependencies()
                 .iter()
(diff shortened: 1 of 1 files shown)


silvi-t (approved):
Tested locally in an e2e scenario — the `kuadrant_filter` span that was previously missing in unsuccessful requests is now back in the traces. LGTM 

**AI review note**: `digest()` (line 197-202) has a similar `Terminate` → `execute_teardown()` path but is not affected by this issue — `pending` is consumed by `apply()` before teardown runs, so there are no lingering span references.

PR #408, opened 2026-07-30
TokenUsageTask returns early if info is present

didierofrivia:
This PR introduces an early check for fields present in order to avoid doing the hard work. It's also useful when external proceses can inject these type of fields and the wasm-shim can be alleviated from doing the body parsing.

## Summary by CodeRabbit

- **Bug Fixes**
  - Response processing now completes immediately when all expected response fields are already available.
  - Prevents unnecessary extraction and waiting for additional response data in these cases.

- **Performance**
  - Reduces processing overhead and improves response handling efficiency when required fields have already been captured.

### DIFF
--- a/crates/kuadrant-filter/src/kuadrant/pipeline/tasks/token_usage.rs
+++ b/crates/kuadrant-filter/src/kuadrant/pipeline/tasks/token_usage.rs
@@ -44,6 +44,16 @@ impl Task for TokenUsageTask {
     fn apply(self: Box<Self>, ctx: &mut ReqRespCtx) -> TaskOutcome {
         let mut task: TokenUsageTask = self.into();
 
+        // Exit if fields already in context response body.
+        if task
+            .expected_response_fields
+            .iter()
+            .all(|field| ctx.response_body.get_value(field).is_some())
+        {
+            return TaskOutcome::Done;
+        }
+
+        // No stored field info found, proceed with response body extraction
         let mut strategy = match task.strategy.take() {
             Some(s) => s,
             None => {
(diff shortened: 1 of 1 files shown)


crates/kuadrant-filter/src/kuadrant/pipeline/tasks/token_usage.rs
```diff
@@ -44,6 +44,16 @@ impl Task for TokenUsageTask {
     fn apply(self: Box<Self>, ctx: &mut ReqRespCtx) -> TaskOutcome {
```
adam-cattermole, line 44: It's worth noting this task is legacy and (very) soon to be removed in favour of the [StoreTask](https://github.com/Kuadrant/wasm-shim/blob/f86c2baa131f51107ef8007cf169d01eadd1ebe2/crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store.rs)

didierofrivia: nice! yup, I reckon this would be just for a PoC I'm working on... Do you reckon that the resulting config from a TokenRateLimit Policy will change the configuration that will result building the Pipeline with the StoreTask or would it be the same?

adam-cattermole: Yeah, so all the bits will be extracted from the responseBodyJson vars and added to a `StoreTask`. So any task leveraging them will wait `Pending` until the `StoreTask` resolves for them

didierofrivia:
Closing this since task will be deprecated soon

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/token_usage.rs
adam-cattermole: See #412

PR #424, opened 2026-09-17
Ordered candidate lists for requestBodyJSON/responseBodyJSON

### LINKED ISSUE #423: Ordered pointer-list extraction: CEL functions, JsonBodyParser, SseBodyParser rewrite, observability
## Summary
Generalizes wasm-shim's `responseBodyJSON`/`requestBodyJSON` CEL functions and body parsers to support an ordered list of JSON Pointer candidates per field, instead of exactly one, and replaces `SseBodyParser`'s OpenAI-only positional heuristic with a provider-agnostic strategy. This is the wasm-shim half of RFC 0024; the CRD/config-generation half is tracked in Kuadrant/kuadrant-operator#2272.

Design: [RFC 0024](https://github.com/Kuadrant/architecture/blob/main/rfcs/0024-token-rate-limit-api-format-compatibility.md), reference-level section "wasm-shim" (all subsections).

### CEL function + static analysis
- [ ] `responseBodyJSON`/`requestBodyJSON` accept either a single string or a list of strings
- [ ] Pointers evaluated in list order against the same parsed body; first resolving value wins
- [ ] "Resolves" = dereferences to a JSON `number`, or a JSON `string` that parses in full as a number; anything else falls through to the next candidate
- [ ] No candidate resolving behaves exactly like today's single-pointer "no such value" case
- [ ] Static AST dependency analysis recognizes list-literal arguments and registers a body dependency for every pointer in the list

### Non-streaming: JsonBodyParser
- [ ] Register one `acutejson` callback per candidate pointer, across all of a field's candidates
- [ ] `finalize_extracted()` picks, per field, the highest-priority pointer that actually matched
- [ ] Evaluation-readiness check requires only one candidate per field to resolve, not all
- [ ] End-of-stream failure check only fails when every candidate for a field is absent

### Streaming: SseBodyParser rewrite
- [ ] Existing "second-to-last event" heuristic removed entirely
- [ ] Quote-bounded leaf-key substrings precomputed per candidate pointer (e.g. `/usage/total_tokens` → `"total_tokens"`)
- [ ] Per dispatched SSE event, cheap `str::contains` scan against precomputed substrings; no match → discarded without JSON parsing
- [ ] On a match, only that event's `data` is JSON-parsed; candidates walked in priority order
- [ ] Later events can overwrite a field's value, but never let a lower-priority candidate overwrite an already-set higher-priority one
- [ ] End-of-stream with no candidate ever resolved fails the task, matching today's single-pointer behavior

### Observability
- [ ] Distinct log line (or metric, if task failures are already metered) for "no candidate resolved for field X", separate from other task-failure causes, covering both parser paths
- [ ] `FailureMode` behavior (default `Allow`, override via `RATELIMIT_REPORT_SERVICE_FAILURE_MODE`) unchanged

## Tests
- [ ] Unit: single-pointer regression, multi-pointer first/second/last match, no match, non-numeric values
- [ ] Unit: JsonBodyParser fallback-candidate match, all-candidates-miss, mixed fields
- [ ] SSE: OpenAI Chat Completions, OpenAI Responses API (named `response.completed`), Anthropic (split `message_start`/`message_delta`, no sentinel), Gemini (usage only in true final chunk, no sentinel)
- [ ] SSE: regression test confirming quote-bounded substring match doesn't false-positive `"input_tokens"` against `"cache_read_input_tokens"`

## Notes for reviewers
This is a substantial internal rewrite of the streaming parser (new strategy, not an incremental patch) — see RFC 0024's Drawbacks section. Suggested review order: CEL function + static analysis → JsonBodyParser → SseBodyParser → observability. PR description should include manual verification steps (e.g. curl a mock SSE endpoint shaped like each provider, confirm `hits_addend` reflects extracted usage).

eguzki:
## What

Generalizes `requestBodyJSON`/`responseBodyJSON` to accept an ordered list of
JSON Pointer candidates plus an optional type hint (RFC 0024). The first
candidate that resolves (and matches the type hint, if given) wins.

```cel
{"total_tokens": responseBodyJSON(["/usage/total_tokens"], "number")}
```

Partially progresses #423: does not close it. Covers: CEL function + static
analysis, non-streaming `JsonBodyParser`, observability. Does **not** cover
the `SseBodyParser` rewrite section (still uses the old "second-to-last
event" heuristic): left for a follow-up PR.

## Deviation from the issue's spec

The issue bakes "resolves = number or numeric string" in as the default for
any list. Instead, that's an opt-in second argument (`"number"`, or
`"string"`/`"bool"`/`"list"`/`"map"`); with no type hint, any non-null value
resolves. Reasoning: these are general-purpose CEL functions, not
token-counting-specific, and there's no good default that doesn't quietly
constrain future non-numeric uses of the list form. The RFC's actual need
still works — the operator would pass `responseBodyJSON([...], "number")`.

## Out of scope

- `SseBodyParser` rewrite (leaf-key scan, priority-vs-recency, streaming test
  matrix) — separate follow-up PR.

### 1. Real end-to-end regression check (docker compose example)

`examples/ratelimit_check_report` already exercises `responseBodyJSON` against a
real Envoy + real LLM inference simulator (`llm-d-inference-sim`, OpenAI-shaped
responses) + real Limitador, using today's plain single-pointer form
(`responseBodyJSON("/usage/total_tokens")`). Run it **unmodified**, so this is an actual regression check of the existing behaviour against this branch's code, not a test of the new syntax (that's what section 2 is for).

1. Build the wasm module (if you haven't already):

```sh
make build   # from repo root; produces target/wasm32-wasip1/debug/wasm_shim.wasm
```

2. Run it:

```sh
cd examples/ratelimit_check_report
make run
```

3. In another terminal, send the non-streaming request from the example's README:

```sh
curl --resolve trlp.example.com:18000:127.0.0.1 "http://trlp.example.com:18000/v1/chat/completions" \
     -d '{"model":"meta-llama/Llama-3.1-8B-Instruct","message":[{"role":"user","content":"Tell me a three sentence bedtime story about a unicorn."}]}'
```

Expect `200 OK`, and `docker compose logs -f limitador` should show a `hits_addend` matching the real token count, since the single-pointer path is untouched.

4. Repeat with the streaming request, exercising `SseBodyParser` instead of `JsonBodyParser` (same expectation):

```sh
curl --resolve sse-streaming.example.com:18000:127.0.0.1 "http://sse-streaming.example.com:18000/v1/chat/completions" \
     -d '{
       "max_tokens": 30,
       "stream": true,
       "stream_options": {
           "include_usage": true
       },
       "messages": [
         {
           "role": "user",
         }
       ]
     }'
```

5. Clean up:

```sh
make clean
```

### 2. Proving the ordered-candidate fallback and type hint

Step 1 only exercises the *first* candidate, since `llm-d-inference-sim` is
OpenAI-shaped only, and neither Anthropic nor Google Gemini can be run
locally at all. `utils/multi_llm_provider` is a dedicated dev/test
environment for this: a mock server (`mock/server.py`) that returns OpenAI-,
Anthropic-, or Gemini-*shaped* responses, chosen per request via an
`X-Mock-Provider` header. See
[`utils/multi_llm_provider/README.md`](utils/multi_llm_provider/README.md)
for the full walkthrough.

```sh
cd utils/multi_llm_provider
make run
```

For each scenario below, run the curl, then check `docker compose logs
limitador | grep hits_addend` for the newest `Report` call.

| Scenario | Command | Expect |
|---|---|---|
| OpenAI shape — 1st candidate (`/usage/total_tokens`) matches | `curl --resolve trlp.example.com:18000:127.0.0.1 "http://trlp.example.com:18000/v1/chat/completions" -H "Content-Type: application/json" -H "X-Mock-Provider: openai" -d '{}'` | `Report` call with `hits_addend: 18` — resolved via the 1st candidate |
| Gemini shape — 1st absent, 2nd (`/usageMetadata/totalTokenCount`) matches | `curl --resolve trlp.example.com:18000:127.0.0.1 "http://trlp.example.com:18000/v1/chat/completions" -H "Content-Type: application/json" -H "X-Mock-Provider: gemini" -d '{}'` | `Report` call with `hits_addend: 18` — resolved via the 2nd candidate |
| Anthropic shape — 1st/2nd absent, 3rd (`/usage/output_tokens`) matches | `curl --resolve trlp.example.com:18000:127.0.0.1 "http://trlp.example.com:18000/v1/chat/completions" -H "Content-Type: application/json" -H "X-Mock-Provider: anthropic" -d '{}'` | `Report` call with `hits_addend: 8` — resolved via the 3rd candidate (an intentional under-count, since Anthropic has no native total) |
| Type hint skips a present-but-non-numeric candidate | `curl --resolve trlp.example.com:18000:127.0.0.1 "http://trlp.example.com:18000/v1/chat/completions" -H "Content-Type: application/json" -H "X-Mock-Provider: ambiguous" -d '{}'` | `Report` call with `hits_addend: 33` — 1st candidate is present but a non-numeric string, so the `number` type hint skips it in favor of the 2nd |
| No candidate resolves — fail-open + observability | `curl --resolve trlp.example.com:18000:127.0.0.1 "http://trlp.example.com:18000/v1/chat/completions" -H "Content-Type: application/json" -H "X-Mock-Provider: unknown" -d '{}'` | `200 OK`, but **no** new `Report` call at all — the `store` action never resolves a value, so the downstream `grpc` report action (which depends on it) never runs; see checks below |

For the `unknown` case specifically, also check:

```sh
docker compose logs envoy | grep -i "no candidate resolved"      # expect a warn log
curl -s http://127.0.0.1:18001/stats/prometheus | grep body_extraction_misses  # expect it incremented by 1
```

**Streaming** — same idea, `-d '{"stream": true}'` instead of `-d '{}'`:

| Scenario | Command | Expect |
|---|---|---|
| OpenAI shape, streaming — resolves normally | `curl --resolve trlp.example.com:18000:127.0.0.1 "http://trlp.example.com:18000/v1/chat/completions" -H "Content-Type: application/json" -H "X-Mock-Provider: openai" -d '{"stream": true}'` | `Report` call with `hits_addend: 18` |
| Anthropic shape, streaming — resolves, but only incidentally | `curl --resolve trlp.example.com:18000:127.0.0.1 "http://trlp.example.com:18000/v1/chat/completions" -H "Content-Type: application/json" -H "X-Mock-Provider: anthropic" -d '{"stream": true}'` | `Report` call with `hits_addend: 8` — `SseBodyParser`'s pre-existing "second-to-last event" heuristic happens to land on the right event for this mock's specific event count; that's a coincidence of the fixture, not a general fix (see `utils/multi_llm_provider/README.md`) |
| Gemini shape, streaming — **known gap, expected to fail** | `curl --resolve trlp.example.com:18000:127.0.0.1 "http://trlp.example.com:18000/v1/chat/completions" -H "Content-Type: application/json" -H "X-Mock-Provider: gemini" -d '{"stream": true}'` | `200 OK`, but **no** `Report` call — the penultimate-event heuristic picks a chunk with no usage data. This is the exact SSE-rewrite gap the follow-up PR closes; seeing it fail here now is the expected, correct result on this branch. (On a checkout without #426/#427, this request hangs instead of returning cleanly — see the note above.) |

For the Gemini streaming case, also check (same as the `unknown` case above):

```sh
docker compose logs envoy | grep -i "no candidate resolved"      # expect a warn log
curl -s http://127.0.0.1:18001/stats/prometheus | grep body_extraction_misses  # expect it incremented by 1
```

```sh
make clean
```

## Summary by CodeRabbit

- **New Features**
  - `requestBodyJSON` and `responseBodyJSON` support ordered JSON Pointer candidates with optional type hints.
  - Candidate resolution supports numbers, strings, booleans, lists and maps, including numeric JSON strings.
  - Added the `kuadrant.body_extraction_misses` metric.
  - Added a multi-provider test environment for OpenAI-, Anthropic- and Gemini-shaped responses.

- **Bug Fixes**
  - Body extraction now tries candidates in order and records unresolved candidates without stopping subsequent actions.

- **Documentation**
  - Documented candidate extraction, streaming limitations, fail-open behaviour and diagnostics.


README.md
```diff
@@ -188,6 +188,8 @@ predicates:
 - requestBodyJSON('/my/value') == 'hello'
 ```
 
+`requestBodyJSON` also accepts an **ordered list** of JSON Pointers instead of a single one, plus an optional type hint, exactly like [`responseBodyJSON`](#responsebodyjsonjson_pointer--json_pointer-type) below (same semantics, applied to the request body).
```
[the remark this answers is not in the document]
eguzki: Fixed

utils/multi_llm_provider/envoy.yaml
```diff
@@ -0,0 +1,147 @@
...
+                                    {
+                                      "type": "grpc",
+                                      "execution": "sequential",
+                                      "var": "report_response",
+                                      "service": "ratelimit-report-service",
+                                      "predicate": "request.path == \"/v1/chat/completions\"",
+                                      "terminal": false,
+                                      "isGuard": false,
+                                      "label": "ratelimit_report",
+                                      "messageBuilder": "envoy.service.ratelimit.v3.RateLimitRequest { domain: \"domain-a\", hits_addend: uint(kuadrant.internal.response.body.total_tokens), descriptors: [ envoy.extensions.common.ratelimit.v3.RateLimitDescriptor { entries: ((request.path == \"/v1/chat/completions\") ? [ envoy.extensions.common.ratelimit.v3.RateLimitDescriptor.Entry { key: \"a\", value: string(1) } ] : []) } ] }",
```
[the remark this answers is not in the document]
eguzki: This is intentionally modeling RFC 0024's own documented interim workaround (no native Anthropic total, output_tokens-only is an acknowledged under-count)

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store.rs
```diff
@@ -192,12 +196,16 @@ impl Task for StoreTask {
...
+                let remaining = parser.remaining_fields();
+                if !remaining.is_empty() {
+                    let candidates: Vec<&[String]> =
+                        remaining.iter().map(|g| g.candidates.as_slice()).collect();
+                    warn!(
+                        "No candidate resolved for field(s) {:?} in '{}': body stream ended \
+                        without a match; the request proceeds and this value is skipped",
+                        candidates, self.path
                     );
+                    METRICS.body_extraction_misses().increment();
```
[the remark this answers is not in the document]
eguzki: Deliberately not counting per-group here. This metric's purpose is "is billing being silently skipped". A per-response concern, not a per-field one.

crates/kuadrant-filter/src/data/cel.rs
```diff
@@ -368,62 +384,249 @@ const RESPONSE_BODY_JSON_FN: &str = "responseBodyJSON";
...
+            RESPONSE_BODY_JSON_DATA,
+            RESPONSE_BODY_JSON_FN,
+            &group.key,
+        ),
+        None => Err(ExecutionError::FunctionError {
+            function: RESPONSE_BODY_JSON_FN.to_string(),
+            message: "Invalid arguments: expected a JSON Pointer string or list of strings, \
+                and an optional type hint string"
+                .to_string(),
+        }),
```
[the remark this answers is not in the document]
eguzki: Deliberately keeping the hard error here. A malformed literal argument (typo'd type hint, oversized list, non-string element) will never work

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store.rs
```diff
@@ -192,12 +196,16 @@ impl Task for StoreTask {
...
                     return TaskOutcome::Failed;
```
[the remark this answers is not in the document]
eguzki: Keeping TaskOutcome::Failed. Traced this through executor.rs/blueprint.rs: store actions are never wrapped in FailureModeTask

crates/kuadrant-filter/src/data/cel.rs
```diff
@@ -368,62 +384,249 @@ const RESPONSE_BODY_JSON_FN: &str = "responseBodyJSON";
...
+    /// candidate to its extracted value, if any.
+    pub(crate) fn resolve<'v>(
+        &self,
+        mut lookup: impl FnMut(&str) -> Option<&'v Value>,
+    ) -> Option<&'v Value> {
+        self.candidates.iter().find_map(|candidate| {
+            let value = lookup(candidate)?;
+            match self.expected {
+                Some(expected) if !expected.matches(value) => None,
+                _ => Some(value),
```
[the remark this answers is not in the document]
eguzki: fixed

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store/sse_body_parser.rs
```diff
@@ -173,22 +174,36 @@ impl BodyParser for SseBodyParser {
...
+            for candidate in &group.candidates {
+                if candidate_values.contains_key(candidate.as_str()) {
+                    continue;
+                }
+                if let Some(value) = json.pointer(candidate) {
+                    let cel_value = match value {
+                        serde_json::Value::String(s) => {
+                            Value::String(std::sync::Arc::new(s.clone()))
+                        }
+                        other => parse_json_scalar(&other.to_string()),
```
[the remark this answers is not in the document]
eguzki: fixed

utils/multi_llm_provider/Makefile
```diff
@@ -0,0 +1,14 @@
+SHELL = /usr/bin/env bash -o pipefail
+.SHELLFLAGS = -ec
+.DEFAULT_GOAL := run
+MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+WORKDIR := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
+DOCKER ?= $(shell which docker 2> /dev/null || echo "docker")
+
+run:
```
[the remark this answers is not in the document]
eguzki: fixed

crates/kuadrant-filter/src/data/cel.rs
```diff
@@ -368,62 +384,254 @@ const RESPONSE_BODY_JSON_FN: &str = "responseBodyJSON";
...
+impl BodyFieldGroup {
+    /// Builds a group from an explicit list of candidates and an optional type hint.
+    /// Exposed as `pub(crate)` (rather than just used internally by the CEL-call
+    /// parsing helpers below) so [`crate::kuadrant::pipeline::tasks::store`] and its
+    /// body parsers can construct/test groups directly, without going through CEL.
+    pub(crate) fn new(candidates: Vec<String>, expected: Option<ExpectedType>) -> Self {
+        let mut key = candidates.join(&BODY_JSON_GROUP_SEP.to_string());
+        if let Some(t) = expected {
+            key.push(BODY_JSON_TYPE_SEP);
+            key.push_str(t.name());
```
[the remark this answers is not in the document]
eguzki: Confirmed — an untyped list candidate containing the raw `BODY_JSON_GROUP_SEP`/`BODY_JSON_TYPE_SEP` control chars could alias the canonical key of an unrelated typed/untyped call, since `dedup_groups` merges purely on that string. Fixed in parse_body_json_args by rejecting any candidate containing either separator (falls back to the existing "unrecognized shape → not registered" path), plus a regression test (`candidate_containing_reserved_separator_is_not_registered`).

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store/json_body_parser.rs
```diff
@@ -262,4 +292,96 @@ mod tests {
...
+    #[test]
+    fn type_hint_with_no_matching_candidate_remains_unresolved() {
+        let field = BodyFieldGroup::new(vec!["/model".to_string()], Some(ExpectedType::Number));
+        let mut parser = JsonBodyParser::new(vec![field.clone()]).unwrap();
+
+        parser.feed(br#"{"model":"gpt-4"}"#).unwrap();
+        parser.finalize().unwrap();
+
+        assert_eq!(parser.remaining_fields(), vec![&field]);
+    }
```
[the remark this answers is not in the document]
eguzki: Dismissing — clippy's `unwrap_used`/`expect_used` restriction lints don't fire inside `#[cfg(test)]` code, which is exactly why `cargo clippy --all-targets --all-features -- -D warnings` already passes clean with 570+ existing `.unwrap()`/`.expect()` calls elsewhere in this crate's tests, despite the crate-level `deny`. Converting these specific tests to `Result`-returning + `?` would just be inconsistent with the rest of the test suite for no enforcement benefit, so leaving as-is.

utils/multi_llm_provider/Makefile
```diff
@@ -0,0 +1,16 @@
...
+
+.PHONY: run clean
+
+run:
+	$(DOCKER) compose -f docker-compose.yaml build mock-llm
+	$(DOCKER) compose -f docker-compose.yaml run start_services
+
+clean:
+	$(DOCKER) compose down --volumes --remove-orphans
+	$(DOCKER) compose -f docker-compose.yaml down --volumes --remove-orphans
```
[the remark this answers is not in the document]
eguzki: Fixed — both `run` and `clean` now pass `-f $(WORKDIR)/docker-compose.yaml` instead of a bare relative `docker-compose.yaml`, and dropped the redundant bare `docker compose down` line (which was the actually dangerous one, having no `-f` at all). This wasn't hit by any current usage in this repo (nothing invokes this Makefile via `make -f` from elsewhere; the README's documented usage is `cd`-ing into the directory first), but scoping explicitly to $(WORKDIR) is cheap and removes the foot-gun for any future/ad-hoc invocation style.

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store/json_body_parser.rs
```diff
@@ -20,14 +21,24 @@ pub(crate) struct JsonBodyParser {
...
+            for candidate in &group.candidates {
+                candidates.insert(candidate.as_str());
+            }
+        }
+
         let mut builder = acutejson::Builder::new();
-        for field in &fields {
-            let field_name = field.clone();
+        for field in candidates {
+            let field_name = field.to_string();
```
adam-cattermole: I think this isn't resulting in the behaviour I was expecting. This now bypasses the benefits of `acutejson`'s streaming, and requires us to read every byte of the body to evaluate that a field is missing - `Done` doesn't fire until we know the candidate isn't there. Prior, we'd exit early as soon as we found the pointer, but now we have to read the entire body to confirm the later entries aren't there even though we could've found a matching candidate

crates/kuadrant-filter/src/data/cel.rs
```diff
@@ -368,62 +384,262 @@ const RESPONSE_BODY_JSON_FN: &str = "responseBodyJSON";
...
+    /// already folds a numeric-looking JSON string into a CEL number before this is
+    /// ever called, which is exactly the "a numeric string resolves as a number"
+    /// behaviour this type hint needs for `Number`.
+    pub(crate) fn matches(self, value: &Value) -> bool {
+        match self {
+            Self::Number => matches!(value, Value::Int(_) | Value::UInt(_) | Value::Float(_)),
+            Self::String => matches!(value, Value::String(_)),
+            Self::Bool => matches!(value, Value::Bool(_)),
+            Self::List => matches!(value, Value::List(_)),
+            Self::Map => matches!(value, Value::Map(_)),
```
adam-cattermole: Are these cases actually reachable? SSE looks like it uses `parse_json_scalar` and `acutejson` looks like it only fires on scalar values too?

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store/json_body_parser.rs
adam-cattermole: I think this also highlights a second problem - before, a potentially truncated or malformed body would still be accepted when the first pointer resolved successfully. Whereas now even if the first candidate resolves and is the expected type, if the body is truncated after that the storetask will fail and dicard the resolved candidate

crates/kuadrant-filter/src/data/cel.rs
```diff
@@ -368,62 +384,262 @@ const RESPONSE_BODY_JSON_FN: &str = "responseBodyJSON";
...
+fn value_as_string_list(v: &Value) -> Option<Vec<&str>> {
+    match v {
+        Value::List(list) => list.iter().map(value_as_string).collect::<Option<Vec<_>>>(),
+        _ => None,
+    }
+}
+
+fn body_json_group_from_values(args: &[Value]) -> Option<BodyFieldGroup> {
+    let (candidates, expected) = parse_body_json_args(args, value_as_string, value_as_string_list)?;
+    Some(BodyFieldGroup::new(candidates, expected))
```
adam-cattermole: nit: I think this rederives on each evaluation instead of the original move arc zero-cost allocation

README.md
```diff
@@ -234,6 +236,25 @@ predicates:
 - responseBodyJSON('/my/value') == 'hello'
 ```
 
+#### `responseBodyJSON([json_pointer, ...], type?)`
+
+Both `requestBodyJSON` and `responseBodyJSON` also accept an **ordered list** of JSON Pointers instead of a single one, to look up a value that may live at different paths depending on which upstream produced the body (e.g. different LLM providers shaping token usage differently). The pointers are evaluated in list order against the same parsed body; the first one that resolves wins, and the rest are never evaluated:
```
adam-cattermole: I think related to https://github.com/Kuadrant/wasm-shim/pull/424/changes#r4072125026 it's actually not entirely true - they're all evaluated regardless - the selection is just ordered

crates/kuadrant-filter/src/data/cel.rs
```diff
@@ -368,62 +384,262 @@ const RESPONSE_BODY_JSON_FN: &str = "responseBodyJSON";
...
+    /// Whether `value` (already parsed out of a matched JSON body fragment) counts as
+    /// "resolved" for this expected type. Note that, for non-streaming bodies,
+    /// [`crate::kuadrant::pipeline::tasks::store::body_parser::parse_json_scalar`]
+    /// already folds a numeric-looking JSON string into a CEL number before this is
+    /// ever called, which is exactly the "a numeric string resolves as a number"
+    /// behaviour this type hint needs for `Number`.
+    pub(crate) fn matches(self, value: &Value) -> bool {
+        match self {
+            Self::Number => matches!(value, Value::Int(_) | Value::UInt(_) | Value::Float(_)),
+            Self::String => matches!(value, Value::String(_)),
```
adam-cattermole: I think `parse_json_scalar` runs before the hint is used - so a string like `"12345"` gets mapped to a `Value::Int` and the hint of `string` causes it to never resolve in a non-SSE case. The SSE case though doesn't go through the same flow and preserves the string so it functions differently in SSE vs non-SSE

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store/json_body_parser.rs
eguzki: Fixed. previously, resolution favored list order. List order still acts as a deterministic tie-break if multiple candidates happen to resolve in the same chunk.

crates/kuadrant-filter/src/data/cel.rs
eguzki: Good catch, not reachable. Compound value extraction doesn't exist today in either backend. Proper compound value support would need real changes to both parsers and is out of scope for this PR.

Dropped List/Map from ExpectedType

crates/kuadrant-filter/src/data/cel.rs
eguzki: this was tricky.  The plain single pointer, no-hint case, now bypasses BodyFieldGroup entirely and reuses the argument's existing `Arc<String>` directly.

README.md
eguzki: Right, not accurate as written. "the first one that resolves wins" is accurate. Changing the "list order" framing to "arrival order, list order only as tie-break". 

Done some rewording.

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store/json_body_parser.rs
```diff
@@ -118,19 +137,36 @@ impl BodyParser for JsonBodyParser {
...
+        // A registered candidate that a group never ends up using (because
+        // some other candidate resolved the group first) still has to stay
+        // registered, since we can't know in advance which one will match;
+        // `acutejson::Status::Done` therefore won't fire until every
+        // registered candidate is seen, not just the ones we still care
+        // about. So resolution is attempted after every chunk rather than
+        // only on `Done`: a group is done as soon as any of its own
+        // candidates resolves, regardless of the other registered paths.
+        let feed_result = parser.feed(chunk);
+        self.finalize_extracted()?;
```
adam-cattermole: I think now that we `finalize_extracted` after every feed there's a chance we've found the json pointer but not read the entirety of it's value yet - the register callback on L49 will have to take into account the `is_complete` before it marks the field as there - just in case a body value is split over multiple chunks

crates/kuadrant-filter/src/data/cel.rs
eguzki: This was even trickier :sweat: . 

Yes, that was a bug. Fixed changing the approach. 

Now, the `body_parser.rs` validates a candidate against its group's hint before producing a `Value`. So, no need to match already produced CEL values with that `matches` function. It fits more naturally, since the same raw candidate can legitimately resolve differently for two groups with different type hints. Only one can resolve. 

Note that 'number' type hint keeps it's documented behavior; it accepts not only a bare JSON number,  but also a JSON string whose content parses in full as a number.

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store/json_body_parser.rs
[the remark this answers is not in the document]
eguzki: Fixed

eguzki: Fixed and added regression test

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store/body_parser.rs
```diff
@@ -32,3 +39,133 @@ pub(super) fn parse_json_scalar(raw: &str) -> Value {
...
+        Some(ExpectedType::Number) => {
+            let value = parse_json_scalar(raw);
+            matches!(value, Value::Int(_) | Value::UInt(_) | Value::Float(_)).then_some(value)
+        }
+        Some(ExpectedType::Bool) => (!is_string)
+            .then(|| parse_json_scalar(raw))
+            .filter(|value| matches!(value, Value::Bool(_))),
+        None => {
+            let value = if is_string {
+                Value::String(Arc::new(raw.to_string()))
```
adam-cattermole: I think this might end up with a change in behaviour right now for expressions without the type hint - if an expression leverages `responseBodyJson('/usage/total_tokens')` and was reading "123" it'd have been mapped to `Int(123)` whereas after this is merged this will be `String("123")` - I'm unsure if this is a realistic use case though and this behaviour should be retained

adam-cattermole:
Was just running through verification as I think we're pretty much there, I didn't get the expected response from the `unknown` case - it does send the `Report` and there's no increment to the metric.. I'll see if I can figure out why

adam-cattermole:
> Was just running through verification as I think we're pretty much there, I didn't get the expected response from the unknown case - it does send the Report and there's no increment to the metric.. I'll see if I can figure out why

Woops, running the wrong stack which ignores the header - makes sense

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store.rs
```diff
@@ -187,17 +191,21 @@ impl Task for StoreTask {
...
-                if !parser.remaining_fields().is_empty() {
-                    let remaining: Vec<&String> = parser.remaining_fields();
-                    error!(
-                        "Body stream ended without finding fields {:?} for '{}'",
-                        remaining, self.path
+                let remaining = parser.remaining_fields();
+                if !remaining.is_empty() {
+                    let candidates: Vec<&[String]> =
+                        remaining.iter().map(|g| g.candidates.as_slice()).collect();
+                    warn!(
```
adam-cattermole: I couldn't decide if this deserves `warn` or should be dropped to `debug` - I think if it doesn't match for a malformed request perhaps `warn` is justified, but if it's misconfigured this'll fire for every request given the default is `warn` as well alongside the metric. Don't mind it being left as is, but perhaps later we'll drop it if it's causing spurious bug reports

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store/body_parser.rs
adam-cattermole: Sure but I guess this is somewhat tangential; having the correct behaviour here prevents issues if that `uint` wrapper was to accidentally be removed. What's the expected behaviour? without type hint should string integers be coerced to integer as before or left as a string?

adam-cattermole: Yeah perhaps you're right, expected behaviour is the new behaviour case? so users can represent something in the body that's an integer as a string and then if they want it as an int it's possible in CEL?

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store.rs
eguzki: You might be right and this might cause too much bothering noise. 

Just to be clear, it only fires when none of a group's candidates match. Thus, for me, if the response could not be parsed, it should be notified per request. It has been explicitly requested like that. Maybe, as you suggest, metric is enough and save log from warn entries. Not sure of that.

adam-cattermole (approved):
Verification looks good, thanks for the changes

PR #427, opened 2026-09-18
Failed tasks do not block sequential

adam-cattermole:
Sequential tasks are blocked by all preceeding tasks from running. If the preceeding tasks fail and are non-terminal, then the pipeline should proceed to try to evaluate the sequential task

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


eguzki (approved):
LGTM

PR #429, opened 2026-09-21
Rewrite SseBodyParser for provider-agnostic streaming extraction

### LINKED ISSUE #423: Ordered pointer-list extraction: CEL functions, JsonBodyParser, SseBodyParser rewrite, observability

eguzki:
## What

Closes #423 

Kuadrant's LLM policies extract token usage from streaming (SSE) responses via `responseBodyJSON`, but different providers put usage data in different places at different points in the stream. Today only OpenAI's shape reliably resolves; Anthropic only produces a partial (output-only) count, and Gemini fails outright, since usage only appears in the true final chunk and the old heuristic doesn't reliably see it.

This makes the ordered-candidate-list form of `responseBodyJSON` (added in #424) work correctly for SSE the same way it already does for plain JSON bodies: a candidate now resolves as soon as a matching value is seen anywhere in the stream, regardless of provider or event count/ordering, rather than depending on where in the stream that event happens to land.

Verified against OpenAI-, Anthropic-, and Gemini-shaped streaming responses via `utils/multi_llm_provider`: all three now report the correct token count end-to-end through Envoy → Limitador, including Gemini, which previously failed to extract usage.

Closes the remaining part of #423 (SSE support for the RFC 0024 candidate-list API).

### 1. Real end-to-end regression check (docker compose example)

`examples/ratelimit_check_report` already exercises `responseBodyJSON` (via
`SseBodyParser`) against a real Envoy + real LLM inference simulator
(`llm-d-inference-sim`, OpenAI-shaped) + real Limitador. Run it
**unmodified**, so this confirms the rewrite didn't regress the one
real-inference streaming path available locally.

1. Build the wasm module (if you haven't already):

```sh
make build   # from repo root; produces target/wasm32-wasip1/debug/wasm_shim.wasm
```

2. Run it:

```sh
cd examples/ratelimit_check_report
make run
```

3. Send the streaming request:

```sh
curl --resolve sse-streaming.example.com:18000:127.0.0.1 "http://sse-streaming.example.com:18000/v1/chat/completions" \
     -d '{
       "max_tokens": 30,
       "stream": true,
       "stream_options": {
           "include_usage": true
       },
       "messages": [
         {
           "role": "user",
         }
       ]
     }'
```

Expect `200 OK`, and `docker compose logs -f limitador` should show a
`Report` call with `hits_addend` matching the real token count — same as
before this PR, since OpenAI's shape was already the one case the old
heuristic handled correctly.

4. Clean up:

```sh
make clean
```

### 2. Proving provider-agnostic streaming resolution

Section 1 only exercises OpenAI's shape, since `llm-d-inference-sim` doesn't
produce Anthropic- or Gemini-shaped output, and neither of those can be run
locally at all (closed, hosted-only APIs). `utils/multi_llm_provider` is a
dedicated dev/test environment for this: a mock server that returns OpenAI-,
Anthropic-, or Gemini-*shaped* SSE streams, chosen per request via an
`X-Mock-Provider` header. See
[`utils/multi_llm_provider/README.md`](utils/multi_llm_provider/README.md)
for the full walkthrough.

```sh
cd utils/multi_llm_provider
make run
```

All three now report the correct token count — this is the actual change
under test: before this PR, Gemini failed outright (no `Report` call at
all) and Anthropic only resolved by coincidence of this mock's specific
event count, not by design.

| Scenario | Command | Expect |
|---|---|---|
| OpenAI shape — usage event resolves as soon as it streams by | `curl --resolve trlp.example.com:18000:127.0.0.1 "http://trlp.example.com:18000/v1/chat/completions" -H "Content-Type: application/json" -H "X-Mock-Provider: openai" -d '{"stream": true}'` | `Report` call with `hits_addend: 18` |
| Anthropic shape — `message_delta`'s `output_tokens` resolves wherever it appears in the stream, not by position | `curl --resolve trlp.example.com:18000:127.0.0.1 "http://trlp.example.com:18000/v1/chat/completions" -H "Content-Type: application/json" -H "X-Mock-Provider: anthropic" -d '{"stream": true}'` | `Report` call with `hits_addend: 8` — the documented under-count via `output_tokens` (Anthropic has no native total), now resolved deterministically rather than by coincidence |
| Gemini shape — `usageMetadata` is only in the true final chunk; **previously failed, now resolves** | `curl --resolve trlp.example.com:18000:127.0.0.1 "http://trlp.example.com:18000/v1/chat/completions" -H "Content-Type: application/json" -H "X-Mock-Provider: gemini" -d '{"stream": true}'` | `Report` call with `hits_addend: 18` — this is the scenario the old "second-to-last event" heuristic could never reliably catch |

Clean up:

```sh
make clean
```

## Summary by CodeRabbit

- **New Features**
  - Streaming responses now detect matching usage data as each event arrives, enabling earlier processing and reporting.
  - JSON candidate lists work more reliably across differing streaming formats from OpenAI, Anthropic and Gemini providers.
  - Higher-priority matching candidates are preserved when multiple candidates are encountered.

- **Bug Fixes**
  - Malformed, unrelated streaming events no longer interrupt processing.

- **Documentation**
  - Updated guidance for streaming candidate-list behaviour and multi-provider token extraction.


eguzki:
@coderabbitai review

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store/sse_body_parser.rs
```diff
@@ -126,59 +126,95 @@ impl EventBuilder {
...
+/// Only works for object-key-shaped pointers, matching the RFC's own scope: a
+/// pointer ending in an array index (e.g. `/items/0`) has no textual key to scan
+/// for at all, since array elements aren't matched by name in the raw JSON text.
+fn leaf_key_substring(pointer: &str) -> Option<String> {
+    let last = pointer.rsplit('/').next()?;
+    if last.is_empty() {
+        return None;
+    }
+    let unescaped = last.replace("~1", "/").replace("~0", "~");
+    Some(format!("\"{unescaped}\""))
```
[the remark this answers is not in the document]
eguzki: Fixed with the minimal, sound version of "use the scan only as an optimisation": leaf_key_substring now returns None for a purely-numeric leaf segment

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store/sse_body_parser.rs
```diff
@@ -201,10 +237,36 @@ impl BodyParser for SseBodyParser {
...
+                group.resolve_indexed(|candidate| candidate_values.get(candidate))
+            {
+                let should_update = match self.resolved_priority.get(&group.key) {
+                    Some(&current_best) => candidate_idx <= current_best,
+                    None => true,
+                };
+                if should_update {
+                    self.extracted.insert(group.key.clone(), value.clone());
+                    self.resolved_priority
+                        .insert(group.key.clone(), candidate_idx);
```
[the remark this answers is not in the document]
eguzki: deliberately keeping this behavior rather than deferring to end-of-stream. Waiting for end-of-stream would block the whole action chain defeating the point of incremental SSE extraction

crates/wasm-shim/tests/streaming.rs
```diff
@@ -156,22 +156,15 @@ fn it_processes_usage_event_across_chunks_until_done() {
...
-        .execute_and_expect(ReturnType::Action(Action::Continue))
-        .unwrap();
-
-    // Second chunk: DONE frame only → captures usage and sends gRPC on end_of_stream
-    let done_chunk = b"data: [DONE]\n\n";
-    let total_len = (usage_chunk.len() + done_chunk.len()) as i32;
-    module
-        .call_proxy_on_response_body(http_context, total_len, true)
-        .expect_get_buffer_bytes(Some(BufferType::HttpResponseBody))
-        .returning(Some(done_chunk))
```
adam-cattermole: This is cool that we can resolve as soon as possible, and the outcome makes sense to me. The one thing it makes me wonder though; if we fire off the `Report`/`Commit` task and end up receiving the response _before_ we've processed the rest of the stream, could we end up with an issue?

If we finish the pipeline and call `resume_http_response()` before we've received eos from the upstream what could happen - I'm hoping that Envoy resumes streaming to downstream and no longer propagates to the filter on the way, but this is behaviour that we couldn't hit before

adam-cattermole:
I was reading that potentially the `totalTokenCount` for Gemini is contained in multiple of the SSE events, where it's cumulative and the last one actually contains the total - I can't find any authoritative confirmation that this is the case though.. it'd mean that the solution for finding the first occurrence and stopping after that would work great for every case apart from Gemini

One of the only things I could find was this [test file](https://github.com/pydantic/pydantic-ai/blob/ad0727bb568580992afebe77351702f8817e5476/tests/models/cassettes/test_google/test_google_stream_usage_is_live_mid_stream.yaml#L28-L39) in another repo. Worth bearing in mind

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store/sse_body_parser.rs
```diff
@@ -194,10 +241,26 @@ impl BodyParser for SseBodyParser {
...
+    }
+}
+
+impl BodyParser for SseBodyParser {
+    fn bytes_consumed(&self) -> usize {
+        0
+    }
+
+    fn finalize(&mut self) -> Result<(), AttributeError> {
+        self.complete = true;
```
adam-cattermole: I think if we say we're complete on finalize and the last event isn't terminated by a blank line `\n\n` we could potentially fail to consume the usage if it's in that last event. I'm unsure how likely the occurrence of this is, so I'll leave this as a nit

adam-cattermole:
I think there's some pre-existing performance enhancements here, as that's unrelated to the changes here I'll leave it as a follow up - multiple chunk copies and allocations, as well as `parse_one_event` allocating a fresh string and copies the remainder on every line

crates/wasm-shim/tests/streaming.rs
eguzki: My opinion, is that in that scenario, calling resume_http_response() before we've received eos from the upstream would stop buffering the response and proceeding further down to the filter chain. I will verify with unittest and one-off e2e manual test with a real envoy. 

Off topic, I wonder why the wasm module is  stopping (`Action::Pause`) on the response body handling when dispatching grpc request. Just let it go, is too late to take any action, just logging.

eguzki: Verified with a unit test plus a live Envoy check.

Unit test (`it_resumes_when_grpc_reply_arrives_before_upstream_finishes_streaming`): forces the report's reply to land before the next body chunk. Confirms wasm-shim's own state machine treats later chunks as a clean pass-through once the pipeline is done. No reprocessing, no extra host calls.

Live check (real Envoy + Limitador, utils/multi_llm_provider, widened the gap after the usage chunk): logs confirm Limitador answered the Report call (hits_addend: 18) ~2s before the final [DONE] chunk arrived, and resume_http_response() fired in between. No Envoy errors/crashes/restarts; client got the complete, correctly-ordered body with no hang. Notably, the filter never even paused here (isGuard: false), so `resume_http_response()` was called on a stream Envoy was never asked to stop, and it still handled it cleanly.

adam-cattermole: > Off topic, I wonder why the wasm module is stopping (Action::Pause) on the response body handling when dispatching grpc request. Just let it go, is too late to take any action, just logging.

There are several reasons we don't drop a pipeline that still has outstanding work to do;

- If we drop the filter before we receive a response, the gRPC response is delivered to envoy without a context that owns it resulting in an error - which isn't necessarily a problem on it's own.
- When it comes to extensions, the behaviour that could be defined to happen in the response phase outgrows our current concerns - we should ensure that we complete the steps we intend to make in the pipeline without dropping it on the floor
- A final reason is that our tracing provider tasks wrap others, and only on teardown do we export traces to the tracing provider. If we rush let envoy kill the filter early all spans that have not yet been sent will be dropped on the floor - there was some discussion of leveraging tracing information for somewhat critical accounting pieces

> Unit test (it_resumes_when_grpc_reply_arrives_before_upstream_finishes_streaming): forces the report's reply to land before the next body chunk. Confirms wasm-shim's own state machine treats later chunks as a clean pass-through once the pipeline is done. No reprocessing, no extra host calls.
>
> Live check (real Envoy + Limitador, utils/multi_llm_provider, widened the gap after the usage chunk): logs confirm Limitador answered the Report call (hits_addend: 18) ~2s before the final [DONE] chunk arrived, and resume_http_response() fired in between. No Envoy errors/crashes/restarts; client got the complete, correctly-ordered body with no hang. Notably, the filter never even paused here (isGuard: false), so resume_http_response() was called on a stream Envoy was never asked to stop, and it still handled it cleanly.

Awesome, thanks for doing this experiment - I have no faith in the proxy-wasm-test-framework of actually producing the same behaviour as envoy, so a real live experiment was needed; was my only real concern so these changes look pretty safe

adam-cattermole (approved):

eguzki:
> I was reading that potentially the `totalTokenCount` for Gemini is contained in multiple of the SSE events, where it's cumulative and the last one actually contains the total - I can't find any authoritative confirmation that this is the case though.. it'd mean that the solution for finding the first occurrence and stopping after that would work great for every case apart from Gemini
> 
> One of the only things I could find was this [test file](https://github.com/pydantic/pydantic-ai/blob/ad0727bb568580992afebe77351702f8817e5476/tests/models/cassettes/test_google/test_google_stream_usage_is_live_mid_stream.yaml#L28-L39) in another repo. Worth bearing in mind

Gemini actually has two separate generation APIs today, and they behave differently here:

- [Interactions API](https://ai.google.dev/gemini-api/docs/interactions-overview?utm_source=gemini) per Google's own docs: 
```
The Interactions API is the best way to build with Gemini models and agents. 
As of June 2026, it is Generally Available and recommended for all new projects. 
While it is now considered legacy, the original generateContent API remains fully supported.
```
 Its streaming produces `step.start`, `step.delta`, `step.stop`, and finally `interaction.completed` events. Detailed text chunks (`step.delta`) carry no usage metadata. Token usage is aggregated and delivered only in the terminal interaction.completed event. For this shape, "first occurrence wins" would be correct, since usage only ever occurs once.
- [Legacy streamGenerateContent](https://ai.google.dev/api/generate-content#v1beta.GenerateContentResponse): Now the legacy path, but still fully supported. Each SSE chunk emits a full GenerateContentResponse object. Every chunk typically includes the usageMetadata object containing running (cumulative) token totals (promptTokenCount, candidatesTokenCount, and totalTokenCount) rather than per-chunk deltas. 

Google's schema reference doesn't document the streaming accumulation behavior itself (I was unable to find it). 

Let's merge this as is and discuss this further, Worth tracking as a follow-up rather than blocking here.

crates/kuadrant-filter/src/kuadrant/pipeline/tasks/store/sse_body_parser.rs
eguzki: As per the format, `data: {...}` followed by nothing isn't a valid event. it's an incomplete event. The stream stopped mid-construction. 

Agree it's a real scenario, but not a bug IMO.

